PR #454, opened 2018-01-17
Use class properties transform to simplify setting the initial state …

wintorez:
## Proposal

Property initializers are detailed in the proposal "ES Class Fields & Static Properties." While an experimental feature that has yet to be ratified, this feature works so well with React that the Facebook team has written about using it internally. Using them will lead to simpler and cleaner code. Please check the following link for further details: https://www.fullstackreact.com/articles/use-property-initializers-for-cleaner-react-components/


wintorez:
Closing this pull request now to resolve some conflicts. Will re-open again. Thanks.

PR #464, opened 2018-01-31
refactor: use class properties transform

wintorez:
## Proposal

Property initializers are detailed in the proposal "ES Class Fields & Static Properties." While an experimental feature that has yet to be ratified, this feature works so well with React that the Facebook team has written about using it internally. Using them will lead to simpler and cleaner code. Please check the following link for further details: https://www.fullstackreact.com/articles/use-property-initializers-for-cleaner-react-components/

Note: this is a re-opening of https://github.com/telusdigital/tds/pull/454

### DIFF
--- a/src/components/ExpandCollapse/Accordion/Accordion.jsx
+++ b/src/components/ExpandCollapse/Accordion/Accordion.jsx
@@ -9,12 +9,8 @@ import Panel from '../Panel/Panel'
  * A specialized `ExpandCollapse` in which only one panel can be open at a time.
  */
 class Accordion extends React.Component {
-  constructor(props) {
-    super(props)
-
-    this.state = {
-      openPanel: props.open,
-    }
+  state = {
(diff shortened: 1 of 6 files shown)


wintorez:
@ryanoglesby08 Thank you for your comment, and also for merging the ticket. I will keep your points in mind in the future.

ahmadnassri:
PR + Issue at the same time is redundant, if the contributor is providing the solution / change at the same time, then there is no need for duplication.

If it's a question / suggestion, then sure, open an issue first, then after discussion add a PR ... 

this was also a topic that was discussed on Slack prior to opening this PR and many people were in favor of it.

style improvements are equally as important as bug fixing or performance changes, else we'd all still be writing ES6 style code.

in this case, there is also a performance impact, when compiling with babel towards a standard Spec, it will certainly improve build times and processing with targeted environment builds with `babel-preset-env`

Issue #267, opened 2017-04-13
[Bug] Forms: dropdowns carets not clickable


twirlingsky:
In the dropdowns for forms the carets are not currently clickable. Also, in Chrome the dropdown information is not styled, it is coming up grey and appears in front of the label. In Firefox the dropdown information comes up below with a white background.

ryanoglesby08:
Testing this on v1.0, seems like this is still a bug. Reopening this Issue until we release a fix.

ryanoglesby08:
Fixed in [v1.0.3](https://github.com/telusdigital/tds/releases/tag/v1.0.3) and [v0.34.4](https://github.com/telusdigital/tds/releases/tag/v0.34.3)

PR #474, opened 2018-02-09
TDS 251 - Make caret in Select component visible and clickable

lzcabrera:
## Description

- Link to related issues: 
  - #267  
  - [TDS-251](https://telusdigital.atlassian.net/browse/TDS-251)
- Description of the problem the contribution solves or link to design issue
  - [ x ] Explanation of how existing TDS code falls short
Select component dropdown caret is not clickable

## Dev Acceptance Criteria

- [ x ] Alignment to approved Sketch file and creative mockups
- [ x ] Documentation is finalized
- [ x ] Code meets AA accessibility guidelines
- [ x ] Code passes automated tests and style checks
- [ x ] New code is unit tested where appropriate (developer should use own judgement)
- [ ] Code passes Pull Request review with 1 approval

### DIFF
--- a/src/components/Select/Select.modules.scss
+++ b/src/components/Select/Select.modules.scss
@@ -24,8 +24,6 @@
   composes: relative from '../Position.modules.scss';
 
   z-index: 2;
-
-  background: transparent;
 }
 
 .iconsPosition {
--- a/src/components/Select/Select.modules.scss
+++ b/src/components/Select/Select.modules.scss
@@ -32,5 +32,6 @@
   composes: absolute centerVertically from '../Position.modules.scss';
   composes: iconRight from '../FormField/FormField.modules.scss';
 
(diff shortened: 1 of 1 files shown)


src/components/Select/Select.modules.scss
```diff
@@ -16,6 +16,8 @@
   &::-ms-expand {
     display: none;
   }
+
+  background-color: transparent;
```
theetrain: This ought to be an explicit colour, in our case `#ffffff` (or the equivalent SCSS variable) since it may render as grey on the AOSP Browser (Android 4.x) or other custom colours applied by users at the OS level.

lzcabrera: I think, I'm just going to remove it because it is not doing anything.
We already have `appearance: none;`.

PR #476, opened 2018-02-12
chore: Introduce a tech snacks file to track small tech tasks

ryanoglesby08:
I wanted to run this by you all for an easy way to track small tech tasks that we want to do. I've used this on other teams and found it valuable.

Tech snacks are small, short, and don't really qualify as a full user story that makes it onto the JIRA board. The word "snack" is purposeful, its not a full meal, just a snack. It should be able to be done in an hour or less, and should be doable by a single developer.

The idea is that when you find a small tech task that is outside of the scope of your current story/task, you can add it to the Tech Snacks list so that we don't lose track of it. When you find that you have a little bit of extra time (end of the day, end of the week, while waiting for a PR review, etc) you can work on one of the outstanding Tech Snacks! Don't forget to remove it from the list when its done.

If we like this idea, we can merge in this PR and then you all can add anything you might know about to the list.

### DIFF
--- a/tech-snacks.md
+++ b/tech-snacks.md
@@ -1,0 +1,8 @@
+## Tech snacks
+
+Tech snacks are small bits of **non-feature** tasks that don't fit nicely into an existing story. They should be able to be
+completed in less than hour. Typical tech snacks focus on refactoring or other cleanup tasks, but could be other things.
+
+If you find yourself with some extra time (at the end of the day, waiting for a PR review, etc), you can work on a tech snack!
+
+* Get rid of "src/scss/settings/\_typography.scss" and incorporate those variables/mixins into the Typography CSS Modules files.


theetrain:
Is this file meant to be updated as we find more 'tech snacks'? I think the 'low hanging fruit' label solves this, and has higher visibility. If existing issues with that label don't feel small enough in scope (less than 1 hour) then perhaps we could re-appropriate the low hanging fruit label?

ryanoglesby08:
> Is this file meant to be updated as we find more 'tech snacks'?

Yes. Add to it and commit it back in. When you complete one, either remove it from the list or we could use strikethroughs to keep history.

> I think the 'low hanging fruit' label solves this, and has higher visibility.

Ahhh, yes, good point about the "low hanging fruit" label. Yes, I suppose that would work too.

My hesitance for that is the I see the Github Issues are more geared towards "real" things such as new features or defects. I would see the tech snacks as even too small for that, and more for the core team.

I feel like it may "clutter" the Issues to be putting these tiny things in there. For example, yesterday I cleaned up some unused docs files, which I would consider a Tech Snack. Is it necessary to create a Github Issue that the support hero is going to have to notice and track and think about and prioritize just for that? Seems overkill.

lzcabrera:
I like the idea of having the low-hanging fruit separate from the core team tech snacks. 
I think we will always have maintenance tasks that the community doesn't need to be aware of.

ryanoglesby08:
@lzcabrera @theetrain 

I'll leave it to you all to decide whether or not to merge this idea. :)

If you all decide to not go this direction, let me know what the strategy is for dealing with these type of internal tech tasks.

PR #493, opened 2018-02-22
fix(clickable): Add default type

marckuo:
- Paired with Enrico
- Added default type: button to clickable
- Updated snapshot tests

### DIFF
--- a/shared/components/Clickable/Clickable.jsx
+++ b/shared/components/Clickable/Clickable.jsx
@@ -24,7 +24,7 @@ Clickable.propTypes = {
   dangerouslyAddClassName: PropTypes.string,
   dangerouslyAddStyle: PropTypes.object,
   children: PropTypes.node.isRequired,
-  type: PropTypes.string,
+  type: PropTypes.oneOf(['submit', 'reset', 'button']),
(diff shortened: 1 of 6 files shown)


shared/components/Clickable/Clickable.jsx
```diff
@@ -23,11 +24,13 @@ Clickable.propTypes = {
   dangerouslyAddClassName: PropTypes.string,
   dangerouslyAddStyle: PropTypes.object,
   children: PropTypes.node.isRequired,
+  type: PropTypes.string,
```
ryanoglesby08: There are actually known, valid values for `type` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button).

PropTypes must use `oneOf` instead of an arbitrary string when there are a distinct set of possible values.

However, I recommend to just remove it as a prop since nothing is actually using that interface. Let's not add the prop until we actually need to. Just hardcode `type="button"` into the JSX and call it a day!

theetrain:
Changed `propTypes` to `oneOf`, that way people can still modify the `type` attribute for StandaloneIcon.

PR #494, opened 2018-02-22
chore: Add gitbook install to the prepare npm script

ryanoglesby08:
Tech snack!

So looks like the "postinstall" npm script is for when users of the package install it. 

"prepare: Run both BEFORE the package is packed and published, and on local npm install without any arguments (See below)"

<https://docs.npmjs.com/misc/scripts>


theetrain:
Does this affect the GitBook deploys? As long as it doesn't install GitBook twice in a row or something, it's okay to merge this.

ryanoglesby08:
Don't think this will effect deploys. Local testing seemed fine. If we find anything wonky its an easy revert.

PR #497, opened 2018-02-26
chore(eslint): Upgrade eslint configs and prettier integration

ryanoglesby08:
Tech snack!

* upgrade eslint-plugin-jest version to fix peer dependency warning
* extend from @telusdigital/eslint-config with recommended syntax
* extend from prettier with recommended syntax for config + plugin


lzcabrera:
Looks good!

PR #498, opened 2018-02-27
TDS-526 - Document independent packages

lzcabrera:
- Updated the styleguidist documentation to reflect the new versioning scheme
- Updated the static documentation to reflect the new versioning scheme (Dev Starter Guide and FAQs)
- Update documentation with instructions on how to upgrade to the independently versioned components, including running the automated transformers ( https://github.com/telusdigital/tds/releases/edit/untagged-3796569693ad29251ea7)


guide/contributing/codebase-overview.md
```diff
@@ -29,22 +29,29 @@ We utilize the following tools for the development, release, and distribution pr
   manager for its speed and deep dependency version locking
 * [Openshift](https://www.openshift.com/) and [Docker](https://www.docker.com/): the CI pipeline is largely
   based on the isomorphic stater kit pipeline, using Docker as the build artifact
+* [Lerna](https://lernajs.io/): A tool for managing the repository that consists of multiple packages.
```
theetrain: We could also add Yarn Workspaces, though it's not a priority.

guide/faq.md
```diff
@@ -10,16 +10,14 @@ Either way, [let us know about it](contact.md) so that we can work with you on t
 
 ## How is TDS versioned?
 
-TDS adheres to [Semantic Versioning](https://semver.org/) so that consumers have a predictable way of managing the dependency.
+As of March 2018, TDS switched to an individually version components and each TDS component is an npm package that adheres to [Semantic Versioning](https://semver.org/) to make it easier for consumers to manage upgrades across independent components.
```
theetrain: > ... `individually-versioned` components...

guide/getting-started/developers.md
```diff
@@ -1,6 +1,6 @@
 # Getting started - developer guide
 
-As a developer, your primary interaction point with TDS is through React components. Though the components are currently distributed in a single bundle, you should conceptualize them as independent, composable components.
+As a developer, your primary interaction point with TDS is through React components. These components are currently are independent, composable components distributed as individual packages.
```
theetrain: > These components `are independent, composable`...

config/styleguide.config.js
```diff
@@ -37,23 +36,34 @@ module.exports = {
...
+    let kebabizeName = name
+      .split(/(?=[A-Z])/)
+      .join('-')
+      .toLowerCase()
+
+    if (name === 'Accordion') {
+      kebabizeName = 'expand-collapse'
+    }
+
+    return `import { ${name} } from '@tds/core-${kebabizeName}'`
```
theetrain: Most of our packages are not named exports, so using deconstruction notation for packages such as `{ FlexGrid }` is invalid. Though I kept a static list of named exports, is there a better way to detect these? https://github.com/telusdigital/tds-codemod/blob/bee9356cd0031457b12c5e92be2d797692a6801a/components.js

I suggest keeping a list of exceptions (namely, ExpandCollapse) to print as named exports.

image

config/styleguide.config.js
```diff
@@ -65,26 +75,19 @@ module.exports = {
       content: path.resolve('docs/intro/welcome.md'),
     },
     {
-      name: 'Foundational elements',
+      name: 'Component Catalogue',
```
theetrain: I suggest `Components & Utilities` or `TDS Packages` since technically this list contains React Component packages and SCSS packages.

guide/contributing/codebase-overview.md
ryanoglesby08: I'd probably just copy the description of lerna straight from their website: "A tool for managing JavaScript projects with multiple packages."

config/styleguide.config.js
ryanoglesby08: Yep, I believe everything is the default export except for ExpandCollapse and Accordion.

guide/contributing/codebase-overview.md
```diff
@@ -29,22 +29,29 @@ We utilize the following tools for the development, release, and distribution pr
...
-    │
+    |  index.cjs.js
+    │  index.es.js
+    |  package.json
+    |  README.md
+    |  rollup.config.js
+    |
     └─── __tests__
         │  ButtonLink.spec.jsx
+        |  ButtonLink.ssr.spec.jsx
```
ryanoglesby08: No need for the SSR spec file here. Not everything has that.

guide/faq.md
```diff
@@ -43,7 +41,7 @@ However, if you are using the ExpandCollpase component, we recommend the use of
 
 The majority of TDS styles are component styles that have been generated by our build tooling to prevent conflicts with application styles. However, there are a small amount of page level styles to establish a common baseline.
 
-Specifically, the `tds.css` stylesheet contains the following page level styles:
+Specifically, the `@tds/core-css-reset/index.css` stylesheet contains the following page level styles:
```
ryanoglesby08: Maybe instead we just link to the CSS Reset package on npm or in the component catalogue and have this information there?

guide/getting-started/developers.md
```diff
@@ -18,13 +18,13 @@ TDS components do not include:
 
 If you are using the [Isomorphic Starter Kit](https://github.com/telusdigital/telus-isomorphic-starter-kit), TDS is already installed and configured. You can jump straight to [using the components](#3-use-tds-components).
 
-To install the latest version:
+To install the latest version of a component:
 
 ```sh
-yarn add @telusdigital/tds
+yarn add @tds/core-component-name
```
ryanoglesby08: Can we make it more apparent that the word "core" is actually part of the package names? How about this?

TDS components follow a standard naming scheme. Core components are prefixed by the word "core".

```
yarn add @tds/core-<component-name>
```

packages/Box/Box.jsx
```diff
@@ -30,6 +30,8 @@ const getBetweenClasses = (scale, inline, desktop) => {
 }
 
 /**
+ * @version 0.0.0-development
```
ryanoglesby08: I think we want `@version 1.0.0` here.... We are going to deploy these docs to prod when we release next, so might as well put version 1.0 now. Find + replace?

packages/Card/Card.jsx
```diff
@@ -7,6 +7,8 @@ import safeRest from '../../shared/utils/safeRest'
 import styles from './Card.modules.scss'
 
 /**
+ * @version 0.0.0-development
```
ryanoglesby08: Does this work right? I was thinking the "@version" had to come AFTER the description?

packages/css-reset/README.md
```diff
@@ -1,3 +1,7 @@
-# TDS Core: CSS Reset
+```css noeditor
+@import '~@tds/core-css-reset/index';`
```
ryanoglesby08: This could also be imported into a .jsx file so webpack can pick it. We want people to use the dist/ one in either case. I think we prefer the jsx one so that the CSS loader can just take it and load it into the bundle. No need to pipe this file through Sass as we have already compiled it to CSS.

```
// in jsx
import '@tds/core-reset/dist/index.css'

// in Sass
import "~@tds/core-reset/dist/index"
```

packages/css-reset/README.md
```diff
@@ -1,3 +1,7 @@
...
+```
+
+Version: 0.0.0-development
 
 Global CSS rules to apply consistent base styles, including a CSS reset, font faces, and root styles.
```
ryanoglesby08: Let's include more info here. Let's include the stuff from the FAQ so it's all in one place.

guide/faq.md
lzcabrera: I'll move this info over to the core-css-reset documentation

packages/Card/Card.jsx
lzcabrera: It works fine. I put it before the description because not all components have a description. 
I like the consistency of having the version number right after the import statement.

packages/css-reset/README.md
lzcabrera: done :)

guide/getting-started/developers.md
```diff
@@ -104,7 +107,7 @@ A list of the available colours and their use cases can be found [in the Design
 ```scss
 // BannerText.scss
 
-@import '~@telusdigital/tds/dist/scss/colours';
```
ryanoglesby08: Realized we should add that the Colours are now a separate package and you will need to install it as a dependency. The getting started guide should call that out, and include a yarn install line for installing it, otherwise you won't be able to to the `@import`!

guide/getting-started/developers.md
```diff
@@ -52,9 +52,9 @@ Add an entry to the `module.rules` array to load CSS. You may also provide other
 }
 ```
 
-### 2. Import the TDS styles
+### 2. Import the TDS css-reset styles
 
-Import the TDS stylesheet into the main entry point of your application. It contains a small amount of page level styles and the pre-compiled components' styles. We use CSS Modules to generate unique class names, which guarantees that TDS component styles will not conflict with your application's class names.
+Import the TDS stylesheet into the main entry point of your application. It contains a small amount of page level styles.
```
ryanoglesby08: Same comment as below with the Colours section. Now that the CSS reset is a separate package, we should tell people to install it!

packages/Box/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tds/core-box",
-  "version": "0.0.0-development",
+  "version": "1.0.0",
```
ryanoglesby08: Ohhhhh, I guess we misunderstood each other. We shouldn't bump the ACTUAL versions of the packages to 1.0 manually. In fact, we should NEVER change the "version" field in our packages manually. That must always be done by running `yarn version`. 

All I meant with my previous comment was to change the JSDoc 
"@version" to 1.0.0. But the packages themselves should stay at 0.0.0-development and we will bump them to 1.0.0 with the release/publish process.

packages/css-reset/README.md
```diff
@@ -1,3 +1,20 @@
-# TDS Core: CSS Reset
+```css noeditor
+// in jsx
+import '@tds/core-reset/dist/index.css'
 
-Global CSS rules to apply consistent base styles, including a CSS reset, font faces, and root styles.
+// in Sass
+import "~@tds/core-reset/dist/index"
```
ryanoglesby08: I know I gave the example..... sorry, just more miscommunication. :(

I think its better NOT to include the Sass way here. We don't want people to do that, so better just to leave it out altogether.

lzcabrera: I removed the sass example and also added a note to use in combination with [ignore-styles](https://www.npmjs.com/package/ignore-styles) npm package

packages/Box/package.json
lzcabrera: Makes sense. I just reverted the commit and added a separate commit to handle the v1.0.0 update on documentation only.

guide/getting-started/developers.md
lzcabrera: We have the last section talking about colours
`### 4. Use TDS colours`

I will add a note to install the dependency.

lzcabrera: And I added a blurb to add the css-reset dependency in 
`### 2. Import the TDS css-reset styles`

guide/faq.md
```diff
@@ -39,19 +37,6 @@ not modify built-in APIs) if it uses a browser API that is not widely supported
...
-
-Specifically, the `tds.css` stylesheet contains the following page level styles:
-
-* **reset**: to standardize default styling of HTML5 elements across browsers, using the popular [Eric Meyer Reset](http://meyerweb.com/eric/tools/css/reset/)
-* **font-face declarations**: to load the TELUS font files
-* **border-box**: to make it easier to size elements by telling the browser to account for any border and padding in the value you specify for width and height
-* **default body styles**: to establish the base font size and colour
-
-Please do not override these global styles. Doing so will likely produce non-deterministic behaviour and UI defects.
-
```
ryanoglesby08: Hmmmm, I'm starting to feel bad about the back and forth.... :(

I did not mean to imply that we ditch the "What global styles..." section altogether. I think it is still a common question, and something users may wonder about. The context of scoping styles is important.

What I meant is to leave this question in the FAQ, keep the part about "The majority of TDS styles are component styles...", and then include a part that says that the only package that DOES set global styles is the CSS Reset package. And include a link to more documentation about that package, either to npm or the component docs.

Does that make sense? Happy to chat about it.

PR #499, opened 2018-02-27
docs(expandcollapse): add details for ontoggle prop

theetrain:
- add some details to the ExpandCollapse `onToggle` prop to add clarity

### DIFF
--- a/packages/ExpandCollapse/ExpandCollapse.jsx
+++ b/packages/ExpandCollapse/ExpandCollapse.jsx
@@ -76,6 +76,7 @@ ExpandCollapse.propTypes = {
   topDivider: PropTypes.bool,
   /**
    * A callback function to be invoked when any panel is opened or closed.
+   * The function can accept a single parameter: `openPanels`.
    *
    * @param {Array} openPanels A list of the currently open panel identifiers.
    */


packages/ExpandCollapse/ExpandCollapse.jsx
```diff
@@ -76,6 +76,7 @@ ExpandCollapse.propTypes = {
   topDivider: PropTypes.bool,
   /**
    * A callback function to be invoked when any panel is opened or closed.
+   * The function can accept a single parameter: `openPanels`.
```
ryanoglesby08: While this seems fine, it really just repeats what is already there... Currently, the docs already list the function arguments...

```
A callback function to be invoked when any panel is opened or closed.
Arguments
openPanels: Array — A list of the currently open panel identifiers.
```

Maybe a wording with less repetition would be `A callback function that will be invoked when any panel is opened or closed with the following arguments...`? And then it lists the arguments as it does now.

Also, this is not the only occurrence of the callback function with arguments... both Accordion and Panel also have this behaviour. It should be consistent.

theetrain:
Closing this due to low impact. Though the representation of function arguments can be improved, the language is already clear.

ryanoglesby08:
@theetrain As that formatting is controlled by styleguidist. The best path towards improving the UI here is to make an Issue/PR on that repo if we care enough. :)

PR #505, opened 2018-03-02
TDS-528 - Fix accessibility issues in Styleguidist

lzcabrera:
We are now loading our own custom AAA colour accessible themes that came from:
- `node_modules/codemirror/theme/ttcn.css`
- `node_modules/highlight.js/styles/docco.css`

Note: As part of this story we found out that the accessible green on athens grey background is not AAA colour accessible with small fonts, so I update the Card colour variants example to not have the green chevron link on grey background Card.

You can now run `yarn dev` and then `yarn test:accessibility` in another tab to see that all 26 nightwatch accessibility assertions pass.

### DIFF
--- a/packages/SelectorCounter/SelectorCounter.md
+++ b/packages/SelectorCounter/SelectorCounter.md
@@ -1,7 +1,10 @@
 ## Minimal usage
 
 ```
-<SelectorCounter />
+<Box between={2}>
+  <label htmlFor="docs_example-1" className="docs_selcounter-label">Simple example</label>
+  <SelectorCounter id="docs_example-1" />
+</Box>
 ```
 
 ## Disabled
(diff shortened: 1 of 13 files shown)


packages/SelectorCounter/SelectorCounter.md
```diff
@@ -1,15 +1,21 @@
 ## Minimal usage
 
 ```
-<SelectorCounter />
+<Box between={2}>
+  <label htmlFor="docs_example-1" className="docs_selcounter-label">Simple example</label>
+  <SelectorCounter id="docs_example-1" />
```
theetrain: Is there a reason why we're using `id` attributes for test selectors and not `data-test`?

lzcabrera: We can't link the label and selector counter input using `data-test-id`, `htmlFor` looks for an id attribute.

ryanoglesby08: @theetrain This is a markdown file, not a test. :)

This id is only used for the example. Because the SelectorCounter does not have a `label` element like our other form fields (Input, Select, etc), it needs to be properly associated with the form element.

theetrain: I see, I totally missed the context here 😲

PR #510, opened 2018-03-06
docs(css-reset): Adjust docs to fix things that were not rendering properly

ryanoglesby08:
* Small update to colours docs too
* Removed instructions for installing packages, as that belongs in getting started docs, not styleguidist docs
* Also updated a Tooltip snapshot...

### DIFF
--- a/packages/css-reset/README.md
+++ b/packages/css-reset/README.md
@@ -1,20 +1,16 @@
-```css noeditor
-// in jsx
-import '@tds/core-reset/dist/index.css'
+```js static
+// App.jsx
+
+import '@tds/core-css-reset/dist/index.css'
 ```
 
 Version: 1.0.0
 
-These stylesheet includes a small amount of page level styles to establish a common baseline.
+This package includes a small amount of page-level styles to establish a common baseline.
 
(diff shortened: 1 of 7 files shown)


packages/css-reset/README.md
```diff
@@ -1,20 +1,16 @@
...
+
+import '@tds/core-css-reset/dist/index.css'
 ```
 
 Version: 1.0.0
 
-These stylesheet includes a small amount of page level styles to establish a common baseline.
-
-Specifically, the index.css stylesheet in this package contains the following page level styles:
+This package includes a small amount of page-level styles to establish a common baseline.
```
lzcabrera: I think we need a semi-colon here:

`comman baseline: `

ryanoglesby08:
@lzcabrera Unless the PR is something really impactful and complicated, I think that if you review and "approve", then you can just go ahead and merge it in right away. No need to wait for other reviewers I think for normal PRs.

If you have comments of things that should be changed, then please mark it as "request changes" instead of approving it so the submitter knows to change something.

@theetrain does that make sense?

Issue #462, opened 2018-01-31
[Radio] text mis-alignment


Berjesty:
## Description

- noticed with TDSv1 radio button's outer circle is not center aligned with the text, specially when we have more than 2 options this becomes more noticeable, can you fix this pls?

## Reproduction Steps

- Click this thing
- Now click this
- Observe this happens

## Reproduction Percentage (if applicable)

- 0/10 times

## Meta

- [ ] TDS version: v1
- [ ] Willing to develop solution
- [ ] Has workaround
- [ ] High impact

## Screenshots

- 
![screen shot 2018-01-31 at 12 50 07 pm](https://user-images.githubusercontent.com/25160096/35639878-977242bc-0689-11e8-85b9-fc7743038f4e.png)

theetrain:
Thanks for the report. If we do not get around to this in the near term, anyone is welcome to work with us in contributing a solution in aligning the radio input's styles with their Sketch designs. It's a low hanging fruit!

lucylist:
hey @Berjesty can you confirm that this is in code and not in DSM? 
looked at the library and it seems vertically centred.

<img width="388" alt="screen shot 2018-02-01 at 4 42 41 pm" src="https://user-images.githubusercontent.com/17018330/35704933-fef15f64-076e-11e8-8703-935e8c11d097.png">

ryanoglesby08:
This is in code. Looking at tds.telus.com you can see the slight misalignment.

Seems like a little bit of pixel pushing to fix.

ryanoglesby08:
Fixed in `@tds/core-radio/1.0.1` and `@tds/core-checkbox/1.0.1`

theetrain:
Also patched in:
- `@telusdigital/tds@0.34.6`
- `@telusdigital/tds@1.0.5`

PR #516, opened 2018-03-09
fix(choice): fix radio button label alignment issue

marcod1419:
#462 - Fixed a small alignment issue with the buttons and their labels. The radio buttons have been moved up by 1px.

**Before**
Before

**After**
After

### DIFF
--- a/shared/components/Choice/Choice.modules.scss
+++ b/shared/components/Choice/Choice.modules.scss
@@ -24,7 +24,7 @@
   justify-content: center;
   cursor: pointer;
 
-  margin-top: 3px;
+  margin-top: 0.125rem;
 
   transition: border-color 0.1s linear, background-color 0.1s linear;
 }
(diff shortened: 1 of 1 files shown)


shared/components/Choice/Choice.modules.scss
```diff
@@ -24,7 +24,7 @@
   justify-content: center;
   cursor: pointer;
 
-  margin-top: 4px;
+  margin-top: 3px;
```
ryanoglesby08: I'd call this a "magic number", meanings its a very random looking number that just happens to work. But why? Why 3 and not 4? Is 3 based on the font size? Is it exactly some percentage of the font size or line height?

Magic numbers are often very confusing and forces us to pixel push all the time to do adjustments.

I suggest taking a little bit of time to see if this can be based on an established variable that we have, or a percentage, or something.

If it is necessary to have a magic number here, lets at least have a comment explaining why its there and what it means.

PR #517, opened 2018-03-09
TDS 326 - Visual Regression

lzcabrera:
As discussed during stand up. Here is what I have working:

`yarn test:2e2`

Should we be moving the command to the individual `package.json` instead?
I have also identified the pattern needed for the scaffolding script.


e2e/config/staging.js
```diff
@@ -0,0 +1,9 @@
+const launchUrl =
+  'http://telus-design-system-docs.s3-website-us-east-1.amazonaws.com/staging/components/index.html'
+const healthCheckUrl =
+  'http://telus-design-system-docs.s3-website-us-east-1.amazonaws.com/staging/components/index.html'
+
+module.exports = {
+  launchUrl,
+  healthCheckUrl,
+}
```
ryanoglesby08: Changing environments seems a bit out of scope for this first implementation, as we are not totally sure how we will using this yet. I suppose having the staging config makes sense as we may want to run visual checks before pushing to prod, but I'm not so sure about the production config.

It seems a bit overkill to have 3 files too, and then use a dynamic `require` to load one of them. I'd probably just have a single `config` object:

```js
const config = {
  development: { ... },
  staging: { ... },
  production: {... }
}

module.exports = config[process.env.APP_ENV || 'development']
```

e2e/globals.js
```diff
@@ -0,0 +1,64 @@
...
+  console.log(`${counter}/20 Health Checking the UI Container...`)
+  request(config.healthCheckUrl, (err, response, body) => {
+    if (!body && counter < 20) {
+      counter += 1
+      setTimeout(healthCheck, 6000, done)
+    } else {
+      done()
+    }
+  })
+}
```
ryanoglesby08: The `healthCheck` seems to be assuming that the e2e tests are also starting up a server, so its waiting for the server to be fully started. I'd recommend removing this part if we don't need it yet.

e2e/page-objects/boxPage.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable no-unused-expressions */
+const { accessibilityScan, visualScan } = require('../tests/steps/common.steps')
+
+const boxPage = {
+  visitPage(browser) {
+    const url = `${browser.launch_url}/#!/Box`
+
+    browser.url(url)
+    this.waitForElementVisible('@header').expect.element('@backToSpa').to.be.visible
```
ryanoglesby08: Why wait for the "backToSpa" element to be visible? Is it not enough to just wait for the "header" to be visible?

ryanoglesby08:
> Should we be moving the command to the individual package.json instead?

Maybe? But would rather not have to duplicate that into every package.json, because then when we want to change something we have to do the same change 40 times!

Instead, we can try to run commands from the root package and scope them by component there.

Here is what I'm imaging would be cool
```sh
yarn test:e2e core-button

# Then, we could integrate lerna. The "lerna updated" command should return modified packages.
# So something like this could pipe the updated package names into the yarn test:e2e command to run the tests on all updated components. :)
# This is similar to how the "husky" precommit stuff works, it passes modified files into prettier/linter
lerna updated | yarn test:e2e
```

ryanoglesby08:
Oh, another request. Seems like you got a lot of good stuff from another repo, I think from Jeff Chang. May be good to include some comments telling future maintainers of this code where the code came from so that we can reference it if needed

e2e/config/staging.js
lzcabrera: updated.

e2e/globals.js
lzcabrera: Not quite sure why it is needed it, but the tests don't run unless we perform the health check immediately after starting the chromedriver.

Either way, I'd prefer to keep it because if I start seeing the healthCheck loop messages on my console I'm quickly reminded that it's most likely because I've forgotten to run `yarn dev` on my other tab.

e2e/page-objects/boxPage.js
lzcabrera: waiting for header is enough. I've removed backToSpa check because it is redundant.

e2e/globals.js
ryanoglesby08: I see. Ok you convinced me!

PR #520, opened 2018-03-13
docs(styleguidist-overrides): Separate styleguidist overrides from custom components

ryanoglesby08:
* Split docs-only components into custom/ and overrides/
* Turn off prettier and linting for the overrides, as they follow the styleguidist conventions, not ours


docs/components/overrides/Editor/Editor.js
```diff
@@ -7,12 +7,12 @@ import 'codemirror/mode/jsx/jsx'
 // We’re explicitly specifying Webpack loaders here so we could skip specifying them in Webpack configuration.
 // That way we could avoid clashes between our loaders and user loaders.
 // eslint-disable-next-line import/no-unresolved, import/no-webpack-loader-syntax
```
theetrain: This, as well as other outstanding `// eslint-disable-*` lines can be removed.

docs/components/overrides/Editor/Editor.js
```diff
@@ -7,12 +7,12 @@ import 'codemirror/mode/jsx/jsx'
...
-require('!!../../../node_modules/react-styleguidist/loaders/style-loader!../../../node_modules/react-styleguidist/loaders/css-loader!../../../node_modules/codemirror/lib/codemirror.css')
+require('!!../../../../node_modules/react-styleguidist/loaders/style-loader!../../../../node_modules/react-styleguidist/loaders/css-loader!../../../../node_modules/codemirror/lib/codemirror.css')
 
 // loading our own (TDS) custom AAA colour accessible ttcn.css theme instead of loading the ttcn theme from the
 // codemirror package mapped to the rsg-codemirror-theme.css webpack alias
 // eslint-disable-next-line import/no-unresolved, import/no-webpack-loader-syntax
```
theetrain: Please remove

docs/components/custom/Markdown/Markdown.js
```diff
@@ -5,20 +5,20 @@ import PropTypes from 'prop-types'
 import { compiler } from 'markdown-to-jsx'
 import mapValues from 'lodash/mapValues'
 // import memoize from 'lodash/memoize';
```
theetrain: This can be removed if not used, also the `eslint-disable` comment at the top of this file.

ryanoglesby08:
> also there are a few errors when rendering Styleguidist

Those are unrelated and were already happening, so I'd rather not fix them as part of this PR.

docs/components/overrides/Editor/Editor.js
ryanoglesby08: I tried to remove this...... but then our editors didn't format correctly. Seems that we still need this.

docs/components/overrides/Editor/Editor.js
ryanoglesby08: I'm leaving these here as this is in docs/overrides, which means it is just overriding a styleguidist component. So it should be left as is so that it stays as similar as possible to the original.

ryanoglesby08:
@theetrain I've addressed the feedback.

Issue #518, opened 2018-03-12
[FlexGrid] Content alignment off by 8px


ah-arch:
## Description

* The content for FlexGrid Columns compared to all other content are:
  - 8px shifted left - For left and center justified columns
  - 8px shifted to the right, when gutterless

## Reproduction Steps

* Visit [TDS FlexGrid example](https://tds.telus.digital/components/index.html#flexgrid): 1.0.0
* Observe *Basic usage* 1st example is aligned 8px off compared to other examples on entire page
* Observe *Distribution* 2nd example is aligned -8px off compared to other examples on entire page

## Screenshots
This is a screenshot of the TDS site, rearranged and highlighted. The red border just shows the Col area. The blue area shows the wrapper container of the FlexGrid, which is where content begins and ends for all other components.

### Current
The content begins either 8px before or after the blue area.
<img width="964" alt="flexgrid - current" src="https://user-images.githubusercontent.com/4450690/37310202-9346fe76-2619-11e8-8272-71e2ef7e0ea3.png">

### Expected
The content is within the blue area, with the right gutter between them.
<img width="967" alt="flexgrid - expected" src="https://user-images.githubusercontent.com/4450690/37310352-05e99d26-261a-11e8-873e-f4d081b64558.png">

## Meta

* TDS version: v1.0.3
* Willing to develop solution: Yes
* Has workaround: Yes
* High impact: ?

## Recommendation

* The FlexGrid Row margin offset to match [Col padding](https://github.com/telusdigital/tds/blob/8d9f4472b5cdd99fa62549f1d68d268cc9df4c25/packages/FlexGrid/Col/Col.modules.scss#L1-L4)
```
.row {
  margin-left: -1rem;
  margin-right: -1rem;
}
```
* Apply [.gutterless](https://github.com/telusdigital/tds/blob/8d9f4472b5cdd99fa62549f1d68d268cc9df4c25/packages/FlexGrid/FlexGrid.modules.scss#L35-L38) only to Col
```
.gutterless { // Col
  padding-left: 0;
  padding-right: 0;
}
```
* Create `.gutterless` for Row
```
.gutterless { // Row
  margin-left: 0;
  margin-right: 0;
}
```

theetrain:
Thanks for reporting, this has been added to our backlog https://telusdigital.atlassian.net/browse/TDS-538

If you wish to see this patch applied soon, you are welcome to contribute!

ah-arch:
[...]

PR #526, opened 2018-03-16
feat(core-flex-grid): fix flexbox gutterless alignment issue and deprecate centre prop

marcod1419:
This pull request addresses issue #518 and also deprecates the `centre` prop. This was done because the fixedWidth prop will also centre the grid, making `centre` redundant.

### DIFF
--- a/packages/FlexGrid/FlexGrid.jsx
+++ b/packages/FlexGrid/FlexGrid.jsx
@@ -18,8 +18,11 @@ import styles from './FlexGrid.modules.scss'
 const FlexGrid = ({ centre, limitWidth, gutter, children, ...rest }) => {
   const gutterStyle = gutter ? undefined : styles.gutterless
 
-  if (centre){
-    deprecate("core-flex-grid", "The centre prop is deprecated due to the limitWidth prop centring the grid on its own. Please remove the centre prop from your grid definition.")
+  if (centre) {
+    deprecate(
+      'core-flex-grid',
+      'The centre prop is deprecated due to the limitWidth prop centring the grid on its own. Please remove the centre prop from your grid definition.'
+    )
   }
 
   const getClasses = () =>
--- a/packages/FlexGrid/FlexGrid.jsx
+++ b/packages/FlexGrid/FlexGrid.jsx
@@ -16,8 +16,6 @@ import styles from './FlexGrid.modules.scss'
  * @version 1.0.0
  */
 const FlexGrid = ({ centre, limitWidth, gutter, children, ...rest }) => {
-  const gutterStyle = gutter ? undefined : styles.gutterless
-
   if (centre) {
     deprecate(
       'core-flex-grid',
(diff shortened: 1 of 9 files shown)


packages/FlexGrid/FlexGrid.modules.scss
```diff
@@ -35,4 +35,5 @@ $max-widths: (
 .gutterless {
   padding-left: 0 !important;
   padding-right: 0 !important;
+  margin: 0 auto !important;
```
ryanoglesby08: @theetrain @marcod1419 

I'm sure you all are aware of the hesitation for using `!important`, so I'm not a huge fan of doing it more. If this is a quick fix for the bug, then thats fine, but I'm fairly worried about overridding the underlying flexboxgrid classes coming from the open source library we use. Seems very dangerous, which is exactly the reason we don't want our consumers to override TDS classes.

Did you all investigate the notes section of [TDS-538](https://telusdigital.atlassian.net/browse/TDS-538)?

I'm worried that these overrides are only a partial bandaid, and we will still run into issues because the flexboxgrid library we use has CSS variables that define the margins and gutters. So even when we override them here, the math may still be wrong in other places. [See the code here](https://github.com/evgenyrodionov/flexboxgrid2/blob/master/src/flexboxgrid2.css#L1-L11).

packages/FlexGrid/FlexGrid.jsx
```diff
@@ -17,6 +18,13 @@ import styles from './FlexGrid.modules.scss'
 const FlexGrid = ({ centre, limitWidth, gutter, children, ...rest }) => {
   const gutterStyle = gutter ? undefined : styles.gutterless
 
+  if (centre) {
+    deprecate(
+      'core-flex-grid',
+      'The centre prop is deprecated due to the limitWidth prop centring the grid on its own. Please remove the centre prop from your grid definition.'
+    )
+  }
+
```
ryanoglesby08: There are no tests for this behaviour/code path. Please add a test that asserts that the `deprecate` function is called when using the `centre` prop.

[See a similar test here](https://github.com/telusdigital/tds/blob/master/packages/Image/__tests__/Image.spec.jsx#L38-L41).

packages/FlexGrid/Row/Row.jsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { Subscriber } from 'react-broadcast'
 
 import { Row as ReactFlexboxGridRow } from 'react-flexbox-grid'
 
+import styles from '../FlexGrid.modules.scss'
+
```
ryanoglesby08: Since we have a row-specific class now, we should create a "Row.modules.scss" file instead and put that class in there. Our convention is for each component to co-locate its styles with its own CSS modules file.

marcod1419: Done!

PR #527, opened 2018-03-17
fix(misc): remove styleguidist errors for markdownheading

theetrain:
Tech Snack!

## Fixes

- MarkdownHeading no longer yields errors in Styleguidist
- Cleaned up tech snacks file
- PropTypes for the `tooltip` prop in the Input and FormField components has changed to `elementType`, which should no longer yield errors when these components are consumed

## Notes

- the `childrenOfType` propType only applies to the `children` prop
- the tooltip errors on Styleguidist still persist since Styleguidist is loading a local copy of Tooltip while Input references a versioned copy of Tooltip - their component signatures therefore differ. This one may be a bit complex to solve. I think we could try tricking Styleguidist to load each component's versioned documentation and exports, or we can contribute a solution since Styleguidist doesn't seem to have documented support for repos using Lerna.

**Before merging** please draft some release notes for the following components, since they've changed according to `lerna updated`

- `@tds/core-input`
- `@tds/core-select`
- `@tds/core-text-area`
- `@tds/core-tooltip`


docs/components/custom/MarkdownHeading/MarkdownHeading.jsx
```diff
@@ -12,7 +12,7 @@ const MarkdownHeading = ({ level, children }) => (
 
 MarkdownHeading.propTypes = {
   level: PropTypes.string.isRequired,
-  children: PropTypes.string.isRequired,
+  children: PropTypes.array.isRequired,
```
ryanoglesby08: I think the more appropriate PropTypes validator here would be `PropTypes.node`. Array doesn't actually make much sense for `children`. I think the error message React is giving is a bit misleading.

ryanoglesby08:
Re: Tooltip Prop type

Something that is interesting is that when you look at the code for `Input` and `FormField`, the `Tooltip` component is not used outside of the PropTypes. So `core-tooltip` is a dependency, but kind of unnecessarily. What if we switched to using the `componentWithName` validator from [Airbnb PropTypes](https://github.com/airbnb/prop-types)? Then we would only use a string, and could remove the core-tooltip dependency from those components

ryanoglesby08:
Re: Tooltip PropTypes

There are other components that use the `childrenOfType` validator besides `Input`, such as [`Select`](https://github.com/telusdigital/tds/blob/master/packages/Select/Select.jsx#L112) and [`Textarea`](https://github.com/telusdigital/tds/blob/master/packages/TextArea/TextArea.jsx#L74). Those should also be updated.

tech-snacks.md
```diff
@@ -6,18 +6,12 @@ completed in less than hour. Typical tech snacks focus on refactoring or other c
...
 * The OpenShift build seems to be running very slow now, seems to be over 20 minutes now for the "build" stage! It didn't used to take this long. Did we change something to make it go slower? Is there any way to speed it up?
 * Inline code blocks (like this one --> `I'm inline code!`) do not render their styles in the styleguide anymore. Fix please. (See an example at <https://tds.telus.com/components/index.html#css-reset> and inspect the words "@font-face".)
-* When running the local styleguide server (`yarn dev`), the following error is in the console: "Warning: Failed prop type: Invalid prop `children` of type `array` supplied to `MarkdownHeading`, expected `string`". Fix please.
```
ryanoglesby08: For context, I was using the strikethrough on purpose so that we could see the history of the tech snacks. Seems that you would rather delete them?

I'm fine either way. Wanted to make sure it was a purposeful decision first though.

theetrain: We decided to delete these snacks as they're completed.

docs/components/custom/MarkdownHeading/__tests__/MarkdownHeading.spec.jsx
```diff
@@ -6,7 +6,9 @@ import MarkdownHeading from '../MarkdownHeading'
 
 describe('MarkdownHeading', () => {
   it('renders', () => {
-    const markdownHeading = shallow(<MarkdownHeading level="h1">Random Heading</MarkdownHeading>)
+    const markdownHeading = shallow(
+      <MarkdownHeading level="h1">{['Random Heading']}</MarkdownHeading>
+    )
```
ryanoglesby08: Please update this test to not pass an array as the child. Using a string is fine. Passing an array as the child doesn't really make sense

packages/Input/Input.jsx
```diff
@@ -1,8 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { childrenOfType } from 'airbnb-prop-types'
+import { componentWithName } from 'airbnb-prop-types'
 
-import Tooltip from '@tds/core-tooltip'
```
ryanoglesby08: If we have removed the only usage of `@tds/core-tooltip` in the Input package, then we should also remove it as a dependency from the package.json of that package. Same goes for Input, Select, TextArea, and FormField.

packages/Input/__tests__/Input.spec.jsx
```diff
@@ -315,6 +315,17 @@ describe('Input', () => {
...
+    it('allows custom tooltip names to be used', () => {
+      const Tool = Tooltip
+
+      const input = mount(<Input label="Some field" tooltip={<Tool>The tooltip</Tool>} />)
+
+      expect(input.find(StandaloneIcon)).toHaveProp(
+        'a11yText',
+        'Reveal additional information about Some field.'
+      )
+    })
```
ryanoglesby08: I am pretty skeptical of the value and readability test...

It seems like it is testing the PropType validation of Tooltip with a pretty obscure usage of renaming the Tooltip component. Also, the assertion that is being made on the StandaloneIcon doesn't match the description of the test: "allows custom tooltip names to be used".

I would recommend to just not have this test at all since it seems to be testing airbnb-prop-types and not our own code.

packages/Select/__tests__/Select.spec.jsx
```diff
@@ -382,6 +383,32 @@ describe('Select', () => {
...
+      const { select } = doMount({
+        tooltip: <Tool>The tooltip</Tool>,
+      })
+
+      expect(select.find(StandaloneIcon)).toHaveProp(
+        'a11yText',
+        'Reveal additional information about The select.'
+      )
+    })
+  })
```
ryanoglesby08: Same comment as on the Input.spec.

packages/TextArea/__tests__/TextArea.spec.jsx
```diff
@@ -347,6 +348,32 @@ describe('TextArea', () => {
...
+      const { textarea } = doMount({
+        tooltip: <Tool>The tooltip</Tool>,
+      })
+
+      expect(textarea.find(StandaloneIcon)).toHaveProp(
+        'a11yText',
+        'Reveal additional information about The textarea.'
+      )
+    })
+  })
```
ryanoglesby08: Same comment as on the Input.spec.

shared/components/FormField/FormField.jsx
```diff
@@ -167,7 +166,7 @@ FormField.propTypes = {
   feedback: PropTypes.oneOf(['success', 'error']),
   error: PropTypes.string,
   helper: PropTypes.oneOfType([PropTypes.func, PropTypes.node]),
-  tooltip: childrenOfType(Tooltip),
+  tooltip: componentWithName('Tooltip'),
```
ryanoglesby08: Since we are changing `FormField`, it is a good idea to give it a package.json and add it as a devDependency to all the components that use it. We did this with `shared/components/Choice`.

This way, any changes to FormField will cause the dependent packages to be re-published.

PR #536, opened 2018-03-22
feat(core-selector-counter): deprecate component

theetrain:
- Deprecates the SelectorCounter component
- Adds notice to documentation and README

**Please rebase and merge** and:

- deprecate this component when publishing to npm
- add deprecation notice to changelog and release notes

image

### DIFF
--- a/packages/SelectorCounter/README.md
+++ b/packages/SelectorCounter/README.md
@@ -1,3 +1,8 @@
 # TDS Core: SelectorCounter
 
 Choose a number.
+
+**This package has been deprecated and is no longer supported.**
+
+The SelectorCounter component is no longer recommended due to it being inaccessible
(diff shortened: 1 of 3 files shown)


ryanoglesby08:
Code seems to look good, but I am expecting the deprecated component name to appear with a strikethrough in the styleguide. Here is an example: https://react-styleguidist.js.org/examples/basic/#wrappedbutton.

Does that not work for us?

screen shot 2018-03-23 at 12 40 29 pm

theetrain:
I wasn't that expecting that strikethrough behaviour. My first guess is it has something to do with us overriding the rsg headers with TDS components.

ryanoglesby08:
@theetrain Ok, what would you like to do? Do you want to investigate the strikethrough behaviour as I suspect is may be a quick fix? Make a new story for adding that back? Or just ignore the strikethrough altogether? I'm fine either way.

theetrain:
Let's avoid Strikethrough text. After discussing with @owestinTelus, I believe strikethrough text does not clearly describe the deprecation.

Issue #465, opened 2018-02-01
Use cases and shortcomings for the Responsive component


theetrain:
## Why

Digital Platform developers traditionally wrote CSS media queries for all their responsive views. It wasn't possible to write unit tests for responsive design, plus writing CSS proved to be time-consuming in general. In response, the `<Responsive />` component was made to tackle these issues.

## What

The `<Responsive />` component was included in TDS as a replacement for CSS media queries. It wraps the [react-media](https://github.com/ReactTraining/react-media/) community component. The key benefits are:

- developers can write unit tests for responsive components
- responsive design is visible within JSX, which maintains context and reduces cognitive strain
- by wrapping other components, you can have responsive JS such as variables that show a different value on desktop and mobile viewports
- performance improvements by letting JS decide which component view to render. This way you can have two or more DOM elements conditionally render rather than having them be present and depend on CSS

There are some known shortcomings with the `<Responsive />` component:

- the starter kit is missing user agent sniffing functionality, which causes issues with server-side rendering. A fix is underway #461 

## Discuss

Has the `<Responsive />` component fallen short for you? Do you have some particular pain points that we can reconcile?

xingbo828:
1. On server side, instead of rendering both desktop and mobile version. I believe render nothing is better. This is the default behaviour of react-media. Somehow it is altered in the `Responsive` component.
2. I don't believe relying on the user-agent is a great idea for the following 2 reasons:
   - TDS should be a stand-alone lib instead of relying on the hosting app to provide additional informations
   - I think user-agent can only gives info about browser type and version, but not the width of the browser. (iPad mobile Safari portraits vs landscape )

ryanoglesby08:
Re: render nothing on server and default behaviour of react-media. -- `Responsive` component does not alter the behaviour of react-media.

Looking at react-media source, it uses a state property of `matches` to determine whether or not to render when there is no `window`. It is set to `true` by default so that it always matches. See here: https://github.com/ReactTraining/react-media/blob/master/modules/Media.js#L21-L27.

The `Responsive` component allows the use of the `defaultMatches` prop of react-media so that consumers can control this behaviour.

We _could_ set `defaultMatches` to false in the `Responsive` component... but not sure if that is the way to go?

ryanoglesby08:
Re: user agent

[...]

PR #537, opened 2018-03-23
feat(core-responsive): leverage sass-mq

marcod1419:
This adds sass-mq to the responsive package. It also refactors some components that used the responsive package to use sass-mq-style media queries. Addresses #465.


packages/DisplayHeading/DisplayHeading.modules.scss
```diff
@@ -1,6 +1,7 @@
 @import '~@tds/core-colours/colours';
 @import '../../shared/scss/typography';
 @import '../../shared/scss/responsive';
+@import '../../packages/Responsive/scss/responsive.scss';
```
ryanoglesby08: This should import responsive.scss from the core-responsive package, so that core-responsive is declared as a dependency. Notice line 1.

It should be something like `@import '~@tds/core-responsive/responsive';`

packages/Heading/Heading.modules.scss
```diff
@@ -1,6 +1,7 @@
 @import '~@tds/core-colours/colours';
 @import '../../shared/scss/typography';
 @import '../../shared/scss/responsive';
+@import '../../packages/Responsive/scss/responsive.scss';
```
ryanoglesby08: This should import responsive.scss from the core-responsive package, so that core-responsive is declared as a dependency. Notice line 1.

It should be something like `@import '~@tds/core-responsive/responsive';`

shared/components/BaseButton/BaseButton.modules.scss
```diff
@@ -1,5 +1,5 @@
 @import '~@tds/core-colours/colours';
-@import '../../scss/responsive';
+@import '../../../packages/Responsive/scss/responsive.scss';
```
ryanoglesby08: This should import responsive.scss from the core-responsive package, so that core-responsive is declared as a dependency. Notice line 1.

It should be something like `@import '~@tds/core-responsive/responsive';`

shared/components/BaseButton/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@tds/shared-base-button",
+  "version": "1.0.0",
+  "private": true,
+  "main": "",
+  "module": ""
+}
```
ryanoglesby08: Nice! :)

packages/Box/__tests__/Box.spec.jsx
```diff
@@ -17,10 +15,6 @@ describe('Box', () => {
     return box.find(box.props().tag)
   }
 
-  beforeEach(() => {
-    mockMatchMedia(575)
-  })
-
   it('renders', () => {
     const box = doMount()
```
ryanoglesby08: The only reason we were doing `mount` in the Box tests was because of the presence of the `Responsive` component, so that we could render its children in the tests. Now that there is no more `Responsive` component in Box, I believe all of these tests can be changed to use `shallow`.

`shallow` rendering is much preferred when we can do it because it is faster than `mount`.

So, these tests should be modified to use `shallow` instead of `mount`.

packages/Button/__tests__/Button.spec.jsx
```diff
@@ -50,20 +48,10 @@ describe('Button', () => {
...
-    })
-
-    it('is inline at medium viewports and above', () => {
-      mockMatchMedia(768)
-
+    it('is being assigned responsive css class', () => {
       const button = doMount()
 
-      expect(button).toHaveClassName('inline')
+      expect(button).toHaveClassName('buttonSizing')
```
ryanoglesby08: There is no logic here right? I think the `buttonSizing` class is always placed on the element, so the presence of that class is tested through the snapshot test I believe.

I think we could just delete this "responsiveness" test.

packages/ButtonLink/__tests__/ButtonLink.spec.jsx
```diff
@@ -55,20 +53,10 @@ describe('ButtonLink', () => {
...
-
-    it('is inline at medium viewports and above', () => {
-      mockMatchMedia(768)
-
-      const link = doMount()
+    it('is being assigned responsive css class', () => {
+      const button = doMount()
 
-      expect(link).toHaveClassName('inline')
+      expect(button).toHaveClassName('buttonSizing')
```
ryanoglesby08: Same comment as above. I think this test could probably be deleted.

packages/DisplayHeading/__tests__/DisplayHeading.spec.jsx
```diff
@@ -3,8 +3,6 @@ import { mount, render } from 'enzyme'
 
 import DisplayHeading from '../DisplayHeading'
 
-import mockMatchMedia from '../../../config/jest/__mocks__/matchMedia'
-
 describe('DisplayHeading', () => {
   const doMount = (props = {}) => {
     const heading = mount(<DisplayHeading {...props}>Great Deals</DisplayHeading>)
```
ryanoglesby08: Same comment as for the Box tests. I think the only reason we used `mount` here was to account for the presence of the `Responsive` component. Without it, we can probably use `shallow` for faster tests.

Please look into switching this to `shallow`.

packages/ExpandCollapse/PanelWrapper/PanelWrapper.modules.scss
```diff
@@ -16,5 +17,21 @@
...
+    }
+  }
+  @include mq($from: md) {
+    span:nth-child(1) {
+      display: initial;
+    }
+    span:nth-child(2) {
+      display: none;
+    }
+  }
```
ryanoglesby08: We don't typically target element selectors in Sass as a convention and best practice (`span:nth-child(1)`), and don't encourage our consumers to do so either. This kind of styles makes it very hard to tell what is going on without having full knowledge of both the markup and the styles.

What do you think about trying it this way instead?

```jsx
<span className={styles.alignSelfFlexEnd}>
  <span className={styles.showFromMd}>
    <Text data-testid="tertiarytext" size="large">{tertiaryText}</Text>
  </span>
  <span className={styles.showUntilMd}>
    <Text data-testid="tertiarytext" size="large">{tertiaryText}</Text>
  </span>
</span>
```

```scss
.showFromMd {
  display: none;

  @include mq($from: md) {
    display: initial;
  }
}

.showUntilMd {
  display: none;

  @include mq($until: md) {
    display: initial;
  }
}
```

shared/components/BaseButton/BaseButton.modules.scss
```diff
@@ -25,19 +25,19 @@ $button-text-color: $color-white;
   transition: background 0.2s;
 }
 
-.fullWidth {
+.buttonSizing {
```
ryanoglesby08: Small comment for the future when using CSS modules. You are already in a file called `BaseButton`, so all class names will be automatically scoped by that. So there is no need to use the word "button" in your class name, as that is implied already. Naming this class just "sizing" would be appropriate.

packages/Button/__tests__/Button.ssr.spec.jsx
```diff
@@ -1,16 +0,0 @@
...
-
-import Button from '../Button'
-
-describe('Button server side rendering', () => {
-  it('uses mobile styles first', () => {
-    const button = render(<Button>Submit</Button>)
-
-    expect(button.hasClass('fullWidth')).toBe(true)
-  })
-})
```
ryanoglesby08: Did all the SSR specs get deleted?

If so, there is a quick cleanup that can be done. These tests were using the `@jest-environment node` syntax to change the environment from jsdom to node for these tests. 

With these tests deleted, I believe this means we can make these changes:
* Delete the "config/jest/setupEnzyme.js" file as this is no longer needed
* Remove the reference to "setupEnzyme.js" from "config/jest.config.js"

packages/ExpandCollapse/PanelWrapper/PanelWrapper.modules.scss
marcod1419: I considered this approach at first but ended up deciding against it because of the extra spans and classes being added in. (I wanted to add as little as possible since this is only a temporary fix) That being said, if you guys tend to not use `nth-child` then I'll switch over to this approach.

ryanoglesby08: Thanks for adding your context.

> because of the extra spans and classes being added in

We have not really been worried about the presence of the "extra" spans, as we would want to measure the impact of that before trying to optimize it. I think the wrapping spans in this case are worth it.

Re: the `nth-child` approach.

One of the big concerns here is that it is completely dependent on the _order_ of the children elements. If someone comes by and rearranges the order of those 2 `<Text>` components for whatever reason, the behaviour would be broken and we wouldn't have any real way of detecting that. The wrapping spans with classes is both more explicit, and less brittle.

packages/Heading/Heading.modules.scss
```diff
@@ -10,63 +11,58 @@
   composes: wordBreak from '../../shared/styles/Typography.modules.scss';
   composes: noSpacing from '../../shared/styles/Spacing.modules.scss';
 
-  @include helvetica-neue-light-45;
+  @include mq($until: md) {
```
theetrain: Perhaps this is personal preference, but I'd rather have a single media query instead of two; like this:

```css
.h1 {
  composes: ...;

  @include helvetica-neue-light-45;
  
  font-size: 1.75rem;
  line-height: 1.29;
  letter-spacing: -1.6px;

  .sup {
    font-size: 1.25rem;
    top: -1em;
  }
  
  @include mq($from: md) {
    font-size: 2.75rem;
    ...
  }
}
```

shared/components/BaseButton/BaseButton.modules.scss
```diff
@@ -25,19 +25,19 @@ $button-text-color: $color-white;
...
   composes: base;
-  display: block;
+  @include mq($until: md) {
```
theetrain: As mentioned above, I prefer using a single media query: 

```css
.sizing {
  composes: base;
  display: block;
  width: 100%;

  @include mq($from: md) {
    display: inline-block;
    width: auto;
    min-width: 180px;
}
```

packages/Button/__tests__/Button.ssr.spec.jsx
marcod1419: The `Box` component still has an SSR test.

ryanoglesby08: Why? It probably doesn't need it anymore? If its all based on media queries, then there is no need for the SSR test.

packages/ExpandCollapse/PanelWrapper/PanelWrapper.modules.scss
marcod1419: Yeah, that's fair enough to want things to be more explicit. I'll be sure to keep that in mind for next time. 😄

marcod1419:
Made the requested changes, let me know if there's anything else!

packages/Button/__tests__/Button.ssr.spec.jsx
marcod1419: Ah, I see, the SSR script just runs a redundant test. I'll remove it then. I also noticed DisplayHeading has one too, so I'll get rid of that as well and clean these up.

marcod1419: Hang on, the tooltip still has an SSR test since it still uses the responsive component. Wouldn't this mean we havet to keep the SSR config stuff in?

ryanoglesby08: Ahhhhh, ok gotcha. Yep makes sense then. 👍

packages/Responsive/Responsive.md
```diff
@@ -65,9 +67,63 @@ initialState = {
...
+  @include mq($until: md, $and: '(orientation: landscape)') {
+    width: 100%;
+    height: 100%;
+  }
+}
+```
+
+#### MQ Parameters
+
+| Parameter     |Values | Effect                                                                                                                         |
```
theetrain: This table looks great, just be sure to left-align each column using `:---` syntax: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables

packages/Responsive/Responsive.md
```diff
@@ -65,9 +67,63 @@ initialState = {
...
+| ------------- |-|-----------------------------------------------------------------------------------------------------------------------------------|
+| `$from`       | `sm`, `md`, `lg`, `xl` |The CSS rules contained within will start at the defined point.                                                                        |
+| `$until`      | `sm`, `md`, `lg`, `xl` |The CSS rules contained within will end at the defined point.                                                                          |
+| `$and`        | Standard CSS Media Query Arguments |Allows for custom arguments like orientation or DPI specifications using the standard CSS syntax. These are encased in quotation marks.|
+
+#### MQ Default Breakpoints
+
+SASS-based media queries will accept the same pre-defined breakpoints as the React component. `sm`, `md`, `lg` and `xl` are all available and break at the same screen widths.
```
theetrain: Change `SASS` to `SCSS`

packages/Responsive/Responsive.md
```diff
@@ -65,9 +67,63 @@ initialState = {
...
+
+Breakpoint key | Width |
```
theetrain: Do these pixel values change depending on whether you use `$from` or `$until`? If so, it may be best to make this a three-column table with the following labels:

| Breakpoint key | Using `$from` | Using `$until` |
| --- | :--- | :--- |
| `sm` | min-width: 576px | max-width: 575px |

shared/components/BaseButton/BaseButton.modules.scss
```diff
@@ -1,5 +1,5 @@
 @import '~@tds/core-colours/colours';
-@import '../../scss/responsive';
+@import '~@tds/core-responsive/responsive';
```
theetrain: Does this require `@tds/core-responsive` to be included in its package.json?

packages/Responsive/Responsive.md
```diff
@@ -1,5 +1,5 @@
 The `Responsive` component is a thin wrapper over the [**react-media**](https://github.com/ReactTraining/react-media) community
-component, which "listens for matches to a CSS media query and renders stuff based on whether the query matches or not."
+component, which "listens for matches to a CSS media query and renders stuff based on whether the query matches or not." It also implements [**sass-mq**](https://github.com/sass-mq/sass-mq) for an easy implementation of responsive behaviour through SASS.
```
ryanoglesby08: We may want to try splitting the docs into 2 sections so that we can clearly separate the responsive Sass stuff from the component docs.

In "styleguide.config.js" we could change the structure like this:

```js
// sections
{
  name: 'Responsiveness',
  content: path.resolve('packages/Responsive/docs/responsive.md'), // Sass docs
  components: path.resolve('packages/Responsive/Responsive.jsx') // component docs
}
```

packages/Responsive/Responsive.md
```diff
@@ -17,6 +17,8 @@ Use the `query` prop if you need to match on [other media features](https://deve
 
 ### Usage
 
+#### React Component
+
```
ryanoglesby08: If we separate the pages with styleguidist we won't need this.

packages/Responsive/Responsive.md
ryanoglesby08: And if we do that, I'd structure the docs like this:

* First, a short, general info to what the responsive package provides.
* Then, show the breakpoint keys and the sizes/queries that they respond to.
* Then, document the Sass stuff
* Then, do the Responsive component and its docs.

packages/Responsive/Responsive.md
```diff
@@ -65,9 +67,63 @@ initialState = {
...
 * [Tooltip component tests](https://github.com/telusdigital/tds/blob/b2108d1074383ba887c5b87a2c3055799937fcd3/packages/Tooltip/__tests__/Tooltip.spec.jsx#L56-L102) and its corresponding [JSX](https://github.com/telusdigital/tds/blob/b2108d1074383ba887c5b87a2c3055799937fcd3/packages/Tooltip/Tooltip.jsx#L81-L108) rendering a particular bubble only when the media query matches
+
+#### SASS Media Queries
+
+SASS-based media queries are available, and are recommended for static sites. This style of responsiveness provides a few benefits like:
+* Prevents sudden scaling when server-side rendered components first appear
+* Prevents the need to wrap Responsive around your component
+  * This in turn removes the requirement for separate CSS classes for different screen sizes that your application must programmatically switch between.
+
+However, using this implementation means losing the ability to programmatically change what is rendered through JSX on resize, so keep the design of your component in mind.
```
ryanoglesby08: I think we don't need to go into detail about why you should use the Sass stuff. I propose that our docs treat the Sass stuff as the default and just gets right into the API, as Sass media queries are well understood.

Instead, lets make it explicit that the Responsive *component* is *not* for general purpose use and it's primary use case is if you want to ensure that a component/element does not appear in the DOM for certain screen sizes. Also mention that the Responsive component may produce DOM mismatches when used with server side rendering. Point people to the workarounds/solutions using defaultMatches and mobile-detect. But reiterate that using the Sass interface is the preferred usage.

The Github Issues about Responsive have a lot of context and background to help provide colour to this explanation.

packages/Responsive/Responsive.md
```diff
@@ -65,9 +67,63 @@ initialState = {
...
+
+To make use of SASS media queries through the Responsive component, first declare the following import at the top of your .scss file.
```
ryanoglesby08: > To make use of SASS media queries through the Responsive component

Clarification: the Sass queries are not coming from the Responsive component. They are coming from the `@tds/core-responsive` package. Small but important distinction.

packages/Responsive/Responsive.md
```diff
@@ -65,9 +67,63 @@ initialState = {
...
+
+```scss
+@import '~@tds/core-responsive/responsive';
+```
+
+Once this is done, you may leverage [**sass-mq**](https://github.com/sass-mq/sass-mq) style media queries to add responsive behaviour to your component.
```
ryanoglesby08: May want to make it more clear that all we are doing is exposing sass-mq directly. "sass-mq style" makes it seem like its similar. Maybe just straight up say, "Importing the responsive Sass file gives you direct access to the sass-mq mixins, configured with our breakpoints"

packages/Responsive/Responsive.md
```diff
@@ -65,9 +67,63 @@ initialState = {
...
+#### MQ Default Breakpoints
+
+SASS-based media queries will accept the same pre-defined breakpoints as the React component. `sm`, `md`, `lg` and `xl` are all available and break at the same screen widths.
+
+Breakpoint key | Width |
+| -------------- | --------|
+| `sm`           | 576px   |
+| `md`           | 768px   |
+| `lg`           | 992px   |
+| `xl`           | 1200px  |
```
ryanoglesby08: I would only put this table in the docs once since its the same between the component and the Sass stuff.

packages/Responsive/Responsive.md
ryanoglesby08: Actually, I think you can just say "Sass". Don't use all caps. https://sass-lang.com/

marcod1419: So we're clear, what is the best time to say Sass and the best time to say SCSS? My bad on the all caps, but I would personally prefer to use "Sass" everywhere because that's the name of the processor.

packages/Responsive/Responsive.md
marcod1419: Nice catch, I'll look into it.

packages/Responsive/Responsive.md
marcod1419: I agree with this, formatting it that way should reduce the need to reiterate the breakpoints. Plus  it'll be easier to push the new Sass stuff as our recommendation.

packages/Responsive/Responsive.md
ryanoglesby08: Yep, say Sass everywhere. No need to ever say SCSS unless you are actually writing a file extension such as `my-styles.scss`

Issue #439, opened 2017-12-08
[Spinner] adds inline styles to body


heathervv:
## Description
- The spinner is adding `overflow: inherit;` to the body as an inline style when added in it's most basic form to a page. Took a look and it _seems_ to be happening from the following code:
```
enableBodyScrolling() {
    document.body.style.overflow = 'inherit'
}
```

## Reproduction Steps
- [ ] Add spinner to page `<Spinner spinning />`
- [ ] Observe inline style on body

## Version Number of TDS
0.29.0

ryanoglesby08:
Core team will be addressing this shortly: https://telusdigital.atlassian.net/browse/TDS-95

ryanoglesby08:
@heathervv I am removing this code from `Spinner`, as well as the `fullScreen` functionality. This will be a breaking change for `Spinner` so it will only be released as new major version on the individual package @tds/core-spinner.

theetrain:
Spinner no longer affects body scroll as of [@tds/core-spinner@2.0.0](https://github.com/telusdigital/tds/releases/tag/%40tds%2Fcore-spinner%402.0.0)

PR #538, opened 2018-03-23
feat(core-spinner): remove fullscreen mode and use CSS modules

ryanoglesby08:
### BREAKING CHANGES
* removal of `wrapperClassNames` prop
* removal of `fullScreen` prop
* forward `rest` props, while disallowing custom styles or CSS classes
* Stop centering the spinner by default, app controls where to put the spinner with layout components or other means

### New prop
* `a11yLabel` sets a title for the SVG so that assistive technology can access it. Defaults to English but app can override as needed.

### Other Spinner changes

* Use CSS modules everywhere
* Lots of style cleanup
* Use correct `accessible-green`
* Use `Text` for the tip to correctly style small text

### Other changes

* Changed the `yarn build` script to use lerna updated to get better resolution of changed files when building

### Github Issues addressed

* #439


ryanoglesby08:
I've been judicious with all commits. Could use Rebase and merge and it should result in a MAJOR version bump for Spinner. Some commits have been marked with BREAKING CHANGE.

PR #539, opened 2018-03-26
feat(build): introduce a publish.js script to scope by packages

ryanoglesby08:
also accept component names on the command line for `yarn build`, `yarn test:e2e`, and `yarn lerna:publish` scripts

examples:
```
yarn publish     # Starts publishing by detecting changed packages with `lerna updated`
yarn publish @tds/core-component     # Starts publishing only @tds/core-component
```

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,39 +1,33 @@
 #!/usr/bin/env node
 
-/* eslint-disable no-console */
+const { spawnSync } = require('child_process')
 
-const { exec, spawnSync } = require('child_process')
+const getUpdatedPackageNames = require('./utils/getUpdatedPackageNames')
+const arrayToGlob = require('./utils/arrayToGlob')
 
-exec('./node_modules/.bin/lerna updated --json', (error, stdout) => {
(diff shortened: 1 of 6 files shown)


scripts/build.js
```diff
@@ -1,39 +1,33 @@
...
+      '--',
+      '$LERNA_ROOT_PATH/scripts/build.sh',
+    ],
+    {
+      stdio: 'inherit',
+    }
+  )
+  spawnSync(
+    './node_modules/.bin/lerna',
+    ['run', '--scope', scopeGlob, '--ignore', '@tds/shared-*', 'build'],
```
theetrain: If you ignore shared components, will those versions not be bumped when there are changes? Or does that not matter since they're private and are always precompiled?

ryanoglesby08: I'll check on this...

ryanoglesby08: Ahhh, this is just the build script, aka rollup. Shared components don't have their own rollup config, so no version bumping is happening here anyway.

The publish.js script does not ignore anything.

PR #563, opened 2018-04-02
feat(docs): read package versions from each package.json file

ryanoglesby08:
* upgrade react-stylguidist to v6.5 to use the new `updateDocs` option
* update scaffolding script accordingly


ryanoglesby08:
NOTE: This PR touches every single package, but only documentation changes, so `lerna updated` will need to be reset. May want to create a `lerna-reset-2` tag after merging this in.

ryanoglesby08:
@marcod1419 conflicts fixed!

PR #565, opened 2018-04-03
chore(dependencies): wip - upgrading rollup and postcss modules

ryanoglesby08:
Pausing on this for now, right now there is a reported bug in rollup-postcss-plugin with importing sass partials.

Will leave this PR here and monitor the issue. If there is no resolution quickly maybe we can help fix the problem.

Here is the blocking issue: <https://github.com/egoist/rollup-plugin-postcss/issues/86>


ryanoglesby08:
Made a PR to fix rollup-plugin-postcss: https://github.com/egoist/rollup-plugin-postcss/pull/90

ryanoglesby08:
This PR is kind of messed up, the only commit that should be here is c53a9ca9c6be0ba58c6fb24ac419218f6afcf3d1.

Enrico is currently working on implemented rollup in tds-community. Once that is done we should be able to apply the same thing to tds-core.

ryanoglesby08:
Replaced by #598

PR #566, opened 2018-04-04
feat(core-flex-grid): add responsive props

marcod1419:
This adds the responsive components xs, sm, md, lg and xl to the col component of flex-grid. It also deprecates the span prop in this component.


ryanoglesby08:
So with `span` being deprecated, I suppose this means this release would be a minor version bump? And then a future release that removes the span prop as the major bump?

packages/FlexGrid/Col/Col.jsx
```diff
@@ -5,42 +5,65 @@ import { Subscriber } from 'react-broadcast'
...
+    <Subscriber channel="flex-grid">
+      {gutter => (
+        <ReactFlexboxGridCol
+          {...removeProps(rest)}
+          xs={xs || span || true}
+          sm={sm || 0}
+          md={md || 0}
+          lg={lg || 0}
+          xl={xl || 0}
+          xsOffset={offset}
```
ryanoglesby08: Maybe we should just spread all of the user supplied props directly into the react-flexbox-grid component? We are not doing anything with them, and our API is essentially the same now as the react-flexbox-grid component. The only "extra" thing we have now is the `span` prop, which can be dealt with separately.

packages/FlexGrid/Col/Col.jsx
```diff
@@ -5,42 +5,65 @@ import { Subscriber } from 'react-broadcast'
...
-        xs={span || true}
-        xsOffset={offset}
-        className={gutter ? styles.padding : styles.gutterless}
-      >
-        {children}
-      </ReactFlexboxGridCol>
-    )}
-  </Subscriber>
-)
+const removeProps = ({ xsOffset, smOffset, mdOffset, lgOffset, xlOffset, ...rest }) =>
```
ryanoglesby08: Would we still want the offsets so that the API is more similar to the react-flexbox-grid component? https://github.com/roylee0704/react-flexbox-grid/blob/master/src/components/Col.js#L13-L17

packages/FlexGrid/Col/Col.jsx
```diff
@@ -5,42 +5,65 @@ import { Subscriber } from 'react-broadcast'
...
+    )
+  }
+
+  return (
+    <Subscriber channel="flex-grid">
+      {gutter => (
+        <ReactFlexboxGridCol
+          {...removeProps(rest)}
+          xs={xs || span || true}
+          sm={sm || 0}
```
ryanoglesby08: `defaultProps` would be the preferred approach here instead of the `||` operator.

However, in this case I'm not sure a default value of `0` is what we want. I think `undefined` is probably more correct. In that the number of columns for those viewports has not been specified, which is different than specifying them to be 0.

packages/FlexGrid/Col/Col.jsx
```diff
@@ -5,42 +5,65 @@ import { Subscriber } from 'react-broadcast'
...
 Col.propTypes = {
   /**
-   * Span the specified number of columns.
+   * @deprecated Span the specified number of columns.
    */
   span: PropTypes.oneOf([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
+  /**
+   * Specify number of columns within the 'xs' breakpoint range. (0px - 575px)
+   */
+  xs: PropTypes.oneOf([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
```
ryanoglesby08: According to react-flexbox-grid, these can also be booleans. https://github.com/roylee0704/react-flexbox-grid/blob/master/src/types.js#L3

packages/FlexGrid/Col/Col.jsx
```diff
@@ -5,42 +5,65 @@ import { Subscriber } from 'react-broadcast'
...
+  /**
+   * Specify number of columns within the 'sm' breakpoint range. (576px - 767px)
+   */
+  sm: PropTypes.oneOf([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
+  /**
+   * Specify number of columns within the 'md' breakpoint range. (768px - 991px)
```
ryanoglesby08: I suggest to remove the pixel values from these comments. Not sure its helpful, and we already have the breakpoints listed elsewhere in the docs. Maybe the FlexGrid docs can just link to where the table is.

packages/FlexGrid/Col/__tests__/Col.spec.jsx
```diff
@@ -46,16 +46,6 @@ describe('Col', () => {
     expect(col).toHaveProp('xsOffset', 2)
   })
```
ryanoglesby08: I think that unless you do mocking for the `deprecate` module we will see `console.log` output in our tests, which is not ideal.

packages/FlexGrid/Col/__tests__/__snapshots__/Col.spec.jsx.snap
```diff
@@ -3,10 +3,14 @@
...
   className="padding"
+  lg={0}
+  md={0}
+  sm={0}
+  xl={0}
   xs={true}
 >
   <div
-    className="col-xs col-xs-offset padding"
+    className="col-xs col-sm-0 col-md-0 col-lg-0 col-xl-0 col-xs-offset padding"
```
ryanoglesby08: Yeah, this snapshot doesn't seem right, with all those `-0` classes.

ryanoglesby08:
This also seems like a good time to update the overall documentation for the FlexGrid as a whole. Particularly how we showcase the grid components, especially in this screenshot.

While cheeky, we don't really use this tone of voice and prefer to have more realistic examples. Showcasing the grid may be an exception, so maybe check out how other grids are documented, such as the react-flexbox-grid, which we use: https://roylee0704.github.io/react-flexbox-grid/ as showcases grids is pretty common.

screen shot 2018-04-04 at 1 34 33 pm

packages/FlexGrid/Col/Col.jsx
marcod1419: We could, but those were not listed in the ACs, so I feel like they might be out of scope.

packages/FlexGrid/Col/Col.jsx
marcod1419: I tried this, and setting these to undefined makes the grid take on generic classes for each breakpoint. For example, if lg is set to undefined, it will take on a generic lg class, meaning that if someone just wanted to set the md prop and have it apply to all greater sizes, it would not work because the generic lg class is overriding the md class. I can look into this more and see if there's a workaround. Good call on the default props.

packages/FlexGrid/Col/Col.jsx
marcod1419: Sure, although maybe then that table should be made to sound more general-purpose rather than just intended for the responsive component in the docs.

ryanoglesby08: Agreed. That table represents the design system breakpoints and every component that uses breakpoints should all be using the same values.

packages/FlexGrid/Col/Col.jsx
```diff
@@ -5,46 +5,110 @@ import { Subscriber } from 'react-broadcast'
...
+        </ReactFlexboxGridCol>
+      )}
+    </Subscriber>
+  )
+}
 
 Col.propTypes = {
   /**
-   * Span the specified number of columns.
+   * @deprecated Span the specified number of columns.
```
theetrain: Please add an alternative prop suggestion.

packages/FlexGrid/Col/Col.jsx
```diff
@@ -5,46 +5,114 @@ import { Subscriber } from 'react-broadcast'
...
-    )}
-  </Subscriber>
-)
+}) => {
+  const flexProps = {}
+
+  if (sm) flexProps.sm = sm
+  if (md) flexProps.md = md
+  if (lg) flexProps.lg = lg
+  if (xl) flexProps.xl = xl
```
ryanoglesby08: I wonder if we making this harder on ourselves here...

Seems like the `PropTypes` definitions below will take care of showing the right interface for the docs, so thats fine.

Would an better solution be to just remove all the responsive and offset props from the component declaration, so that they all get captured by `rest`, and then just do the normal `...safeRest(rest)` to pass everything directly to the underlying react-flexbox-grid component? Then all of this logic with `flexProps` can go away. I think you would also remove the `defaultProps` stuff for all the react-flexbox-grid props too so that exactly what the user specifies will just get forwarded on.

If the linter complains, I'd say in this case its ok to disable the linter check for those props because we have a good reason.

packages/FlexGrid/Col/Col.jsx
```diff
@@ -5,46 +5,114 @@ import { Subscriber } from 'react-broadcast'
...
+import { deprecate } from '../../../shared/utils/warn'
 import styles from './Col.modules.scss'
 
-const removeProps = ({
-  xs,
-  sm,
-  md,
-  lg,
-  xl,
+const removeProps = ({ ...rest }) => safeRest(rest)
```
ryanoglesby08: This can be inlined. The `removeProps` function is now redundant. Just use safeRest directly

packages/FlexGrid/Col/__tests__/__snapshots__/Col.spec.jsx.snap
```diff
@@ -6,7 +6,7 @@ exports[`Col renders 1`] = `
   xs={true}
 >
   <div
-    className="col-xs col-xs-offset padding"
+    className="col-xs col-xs-offset col-sm-offset col-md-offset col-lg-offset col-xl-offset padding"
```
ryanoglesby08: All of these extra offset classes for each breakpoint don't seem right.......

packages/FlexGrid/Col/Col.jsx
marcod1419: I tried this, and weirdly, React will spit out this error: `Error: A cross-origin error was thrown. React doesn't have access to the actual error object in development. See https://fb.me/react-crossorigin-error for more information.` Which is both strange and not very helpful.

ryanoglesby08: I think we can fix that, probably just some syntax error or something

packages/FlexGrid/Col/Col.jsx
```diff
@@ -5,52 +5,111 @@ import { Subscriber } from 'react-broadcast'
...
+    deprecate(
+      'core-flex-grid',
+      `The span prop is deprecated due to the addition of the new responsive props.
+      Replace span in your Col definition with xs for identical functionality.`
+    )
+  }
 
+  const props = rest
+
+  if (offset && !props.xsOffset) props.xsOffset = offset
```
ryanoglesby08: After creating the `props` variable and adding `xsOffset` to it here, you don't do anything with it....  is the intention to pass `props` into `safeRest` now?

theetrain: Yes, good catch

theetrain: Good catch, I'll make that change. Turns out I didn't notice it due to the behaviour of object copying:

image

marcod1419:
@ryanoglesby08 Working on getting those examples updated now.

packages/FlexGrid/Col/Col.jsx
ryanoglesby08: haha yep! I did the same exact thing to check it yesterday too!

If you want to copy an object without effecting the original, the best way is:

```js
const copyOfObj = Object.assign({}, obj)

// or using the spread syntax
const copyOfObj = {...obj}
```

packages/FlexGrid/Col/Col.jsx
```diff
@@ -5,52 +5,111 @@ import { Subscriber } from 'react-broadcast'
...
+  }
+  if (span) {
+    deprecate(
+      'core-flex-grid',
+      `The span prop is deprecated due to the addition of the new responsive props.
+      Replace span in your Col definition with xs for identical functionality.`
+    )
+  }
 
+  const props = rest
```
ryanoglesby08: This is still causing a modification of `rest` though, which is not ideal. Best to make a copy so that the incoming object is not modified accidentally.

```js
const props = {...rest}
```

marcod1419:
Just committed some docs updates, let me know what you guys think and if there's anything I can improve/change!

packages/FlexGrid/Row/Row.md
```diff
@@ -4,21 +4,50 @@ The `FlexGrid.Row` component can be used for Column alignment and distribution.
...
+    </FlexGrid.Row>
+  </Box>
+
+  <Box inset={3}>
+    <FlexGrid.Row horizontalAlign="center" verticalAlign="bottom">
+      <FlexGrid.Col xs={12}>
+        <Button>Submit</Button>
+      </FlexGrid.Col>
+    </FlexGrid.Row>
+  </Box>
```
ryanoglesby08: Honestly I think this is getting a bit overly creative. :) I think the extra components (`Input`, `Button`, `Checkbox`) will obscure the purpose of these examples, which is to give readers and understanding of the grid props and behaviour.

I think using simple language or just solid color in the grid content is superior, as we have done in previous examples. Check out some examples of well documented grids here:

* https://getbootstrap.com/docs/4.0/layout/grid/
* https://semantic-ui.com/collections/grid.html
* https://roylee0704.github.io/react-flexbox-grid/

packages/FlexGrid/Row/Row.md
```diff
@@ -27,19 +56,31 @@ The `FlexGrid.Row` component can be used for Column alignment and distribution.
...
-    <FlexGrid.Col span={3}>Take care</FlexGrid.Col>
+    <FlexGrid.Col xs={4} sm={3}>
+      <Input label="Between 1" />
+    </FlexGrid.Col>
+    <FlexGrid.Col xs={4} sm={3}>
+      <Input label="Between 2" />
+    </FlexGrid.Col>
+    <FlexGrid.Col xs={4} sm={3}>
+      <Input label="Between 3" />
+    </FlexGrid.Col>
```
ryanoglesby08: Perhaps we could combine the two `distribute` examples into one example with 2 rows? Then it will take up less space and be contained.

PR #567, opened 2018-04-04
fix(core-notification): make error text meet WCAG AAA compliance

marcod1419:


packages/Notification/Notification.jsx
```diff
@@ -32,7 +32,7 @@ const renderContent = (variant, children) => {
 
   if (variant === 'error') {
     return (
-      <ColoredTextProvider colorClassName={messagingStyles.errorText}>
+      <ColoredTextProvider colorClassName={messagingStyles.errorTextNotification}>
```
ryanoglesby08: The `ColoredTextProvider` was invented as a sneaky way of changing text color without changing the public API of the Typography components.

If the Notification is now just going to use the default text color, then there is no more need for the ColoredTextProvider component in this case, and you maybe can just remove its use from Notification altogether.

`color-shark` is just the normal `color-text`. https://github.com/telusdigital/tds/blob/master/packages/colours/_colours.scss#L21

packages/Notification/Notification.jsx
```diff
@@ -30,14 +28,6 @@ const renderIcon = icon => <DecorativeIcon symbol={icon.symbol} variant={icon.co
 const renderContent = (variant, children) => {
```
ryanoglesby08: The `content` variable, and actually the entire `renderContent` function, is not necessary anymore. I'd recommend removing it and inlining this into the main component.

packages/Notification/__tests__/Notification.spec.jsx
```diff
@@ -54,11 +53,7 @@ describe('<Notification />', () => {
     it('bolds and colors the content', () => {
```
ryanoglesby08: This test needs a slight name change, and there is no more color change.

ryanoglesby08:
@marcod1419 I realized something else about this one. If the color contrast here was not up to our standards, how come our e2e accessibility test is passing? Are we not configuring aXe properly to catch this level of color contrast?

It would be great if you could see if you can get a failing accessibility test on this component first, and then fix it with the text color change.

marcod1419:
@ryanoglesby08 Just ran the e2e tests and they're passing. Looks like aXe only tests up to WCAG AA. Not seeing anything in their docs for enabling AAA testing.

ryanoglesby08:
Ahhh, ok then. I was thinking that may be the case. Alright, I guess there's nothing we can do then. Are there any issues or requests for AAA settings in aXe core? May be good to do a quick look?

ryanoglesby08:
@marcod1419 @theetrain 

Quick question: is this a fix, feature, or breaking change?

While the API hasn't changed, the style has been visibly altered, which may be surprising to consumers...

PR #576, opened 2018-04-11
chore(tech-snacks): add git lfs snack

theetrain:
- Add Git LFS snack to speed up fetches


theetrain:
Thanks, I added a rationale for this snack. Please **squash and merge** if approved.

PR #577, opened 2018-04-11
chore(docs): full width component playgrounds

ryanoglesby08:
This is a WIP. What do you all think?

Adding the `docs_full-width-playground` class to component playgrounds will cause it to "break out" of it's parent container and display the component using the full browser width. This will allow us to better showcase full width components such as `Notification` or `FlexGrid`.

It is not active when the sidebar is being shown:

screen shot 2018-04-11 at 5 22 22 pm

And it is active when the sidebar is hidden:

screen shot 2018-04-11 at 5 22 37 pm

I briefly attempted other strategies, but this one seemed to be the most straightforward one, even though it does involve some fancy/potentially brittle CSS.

### DIFF
--- a/docs/components/overrides/StyleGuide/StyleGuideRenderer.js
+++ b/docs/components/overrides/StyleGuide/StyleGuideRenderer.js
@@ -5,6 +5,8 @@ import cx from 'classnames'
 import Logo from 'rsg-components/Logo'
 import Markdown from 'rsg-components/Markdown'
 
+import FlexGrid from '../../../../packages/FlexGrid/FlexGrid'
+
 const styles = ({ color, fontFamily, fontSize, sidebarWidth, mq, space, maxWidth }) => ({
   root: {
     backgroundColor: color.baseBackground,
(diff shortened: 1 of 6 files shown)


donnavitan:
Is it possible to have full-width background/image while keeping the content aligned with the rest of the page?

fullwidth

lucylist:
hey @ryanoglesby08 maybe i'm not fully understanding how the current notifications are built now. maybe you guys can explain it to me? I know the examples are clear at the moment on our catalogue.. but.. hoping this live example is how they are currently built? 
screen shot 2018-04-12 at 9 44 43 am

ryanoglesby08:
@donnavitan 

> Is it possible to have full-width background/image while keeping the content aligned with the rest of the page?

The Notification aligns itself with the TDS grid settings, with sets a max width of 1200 px on large viewports.

The content of the component catalogue currently does NOT respect the TDS grid settings, as it is generated by an external library. I can look into changing the width settings of the component catalogue styles to see if we can bring it in line with TDS settings.

ryanoglesby08:
>  It would seem the only way to hide the sidebar is by isolating a single component example.

Styleguidist actually has a config option for this now: https://react-styleguidist.js.org/docs/configuration.html#showsidebar
> Toggle sidebar visibility. Sidebar will be hidden when opening components or examples in isolation mode even if this value is set to true. When set to false, sidebar will always be hidden.

But, if we set that to false there is NEVER a sidebar, which is not good for navigation.

We could potentially build a way to open/close the sidebar on demand, but I didn't look into that yet.

PR #578, opened 2018-04-11
chore(docs): rename repo links to tds-core

ryanoglesby08:
Immediately after merging this PR, the repo name must be changed using the Github UI. https://github.com/telusdigital/tds/settings

I've left all the package.json and changelogs out, Github will set up a redirect so I don't think its worth it to make such a large change and mess with lerna.


ryanoglesby08:
Tasks for QA:

* merge the PR
* Use the Github UI to rename the repo
* Immediately communicate to other TDS devs so we can all rename our git remotes
* Deploy docs to staging and prod

PR #579, opened 2018-04-12
chore(build): improve the prepr script

ryanoglesby08:
...by using the lerna publish command directly

Ok, using a new strategy here by calling a lerna command directly. It uses a non-public interface though, so may be a bit brittle, but I think its likely the most straightforward solution.

Now the output will look like this:

```
yarn run v1.5.1
$ node scripts/prePr.js --conventional-commits --yes
lerna info Comparing with lerna-reset-3.

Changes:
 - @tds/core-flex-grid: 1.1.0 => 1.2.0
 - @tds/core-notification: 1.0.2 => 1.1.0

lerna info auto-confirmed

🤔 If this is not what you expected, ensure that your commit messages follow the Conventional Commits specification: https://conventionalcommits.org.

🚀 Otherwise, paste the entire output of this task into the body of your Pull Request so that a maintainer can verify before merging/publishing.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "lint": "npm-run-all --parallel lint:*",
     "precommit": "lint-staged && yarn test",
     "prepare": "yarn gitbook:install",
-    "prepr": "yarn test:e2e && node scripts/prePr.js",
+    "prepr": "yarn test:e2e && node scripts/prePr.js --conventional-commits --yes",
     "prepush": "npm-run-all --parallel build test build-docs:*",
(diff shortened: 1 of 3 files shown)


ryanoglesby08:
> Small thought, would it be better to include this in prepush so that we get the version readout right after pushing (since we don't have CI running on pull requests and yarn prepr is a manual human task)?

Maybe.... you can push multiple times before making a pr though. Is it useful to see this every time? And to run the e2e tests on every push? I'm not sure. 🤷‍♂️

PR #580, opened 2018-04-12
chore(docs): delete s3 files before deploying wip

theetrain:
Trying out a way to delete files from s3 before deploying by listing files on the bucket, removing them, and then pushing new files.


theetrain:
This isn't a priority at this time.

PR #581, opened 2018-04-13
Improve e2e workflow

marcod1419:
Adds an ability to go through each failing screenshot and update its baseline. The ability to update all screenshots automatically by using a command line argument is not featured in this PR, but will be added shortly.

### DIFF
--- a/e2e/assertions/compareScreenshot.js
+++ b/e2e/assertions/compareScreenshot.js
@@ -3,7 +3,6 @@ const resemble = require('node-resemble-js')
 const fs = require('fs')
 const path = require('path')
 const readline = require('readline')
-const { spawnSync } = require('child_process')
 
 const { tolerance } = require('../config')
 const { getVisualRegressionFolders } = require('../utils')
@@ -18,23 +17,26 @@ const ensureBaselinePhotoExists = (baselinePath, resultPath) => {
 }
 
 const update = (baselinePath, resultPath, callback, data) => {
-  const read = readline.createInterface({
-    input: process.stdin,
-    output: process.stdout,
-  })
-
-  read.question('Would you like to update the baseline for this screenshot? (y/n) ', answer => {
-    if (answer.toLowerCase() === 'y' || answer.toLowerCase() === 'yes') {
-      console.log('Generating new baseline screenshot...')
-      spawnSync('rm', [baselinePath], {
-        stdio: 'inherit',
-      })
-      fs.writeFileSync(baselinePath, fs.readFileSync(resultPath))
-    }
-    read.close()
+  let updateMessage
+  if (process.env.UPDATE_SCREENSHOTS === 'true') {
+    const read = readline.createInterface({
+      input: process.stdin,
+      output: process.stdout,
+    })
+    read.question('Would you like to update the baseline for this screenshot? (y/n) ', answer => {
+      if (answer.toLowerCase() === 'y' || answer.toLowerCase() === 'yes') {
+        console.log('Generating new baseline screenshot...')
+        fs.writeFileSync(baselinePath, fs.readFileSync(resultPath))
+      }
+      read.close()
+      callback(data)
+    })
+  } else if (process.env.UPDATE_ALL_SCREENSHOTS === 'true') {
+    console.log('Generating new baseline screenshot...')
+    fs.writeFileSync(baselinePath, fs.readFileSync(resultPath))
     callback(data)
-  })
-  return 'User was prompted to update screenshot.'
+  }
+  return updateMessage
 }
 
 exports.assertion = function(componentName, fileName) {
@@ -66,13 +68,16 @@ exports.assertion = function(componentName, fileName) {
       // .ignoreAntialiasing()
       // .ignoreColors()
       .onComplete(data => {
-        if (Number(data.misMatchPercentage) > 0.01 && process.env.UPDATE_SCREENSHOTS === 'true') {
+        if (
+          Number(data.misMatchPercentage) > 0.01 &&
+          (process.env.UPDATE_SCREENSHOTS === 'true' ||
+            process.env.UPDATE_ALL_SCREENSHOTS === 'true')
+        ) {
           this.message = update(baselinePath, resultPath, callback, data)
         } else {
           callback(data)
         }
       })
-
     return this
   }
 
@@ -85,12 +90,22 @@ exports.assertion = function(componentName, fileName) {
   }
 
   this.pass = function pass(value) {
-    const pass = value <= this.expected
+    let pass = value <= this.expected
 
-    if (process.env.UPDATE_SCREENSHOTS === 'false' || pass) {
+    if (
+      (process.env.UPDATE_SCREENSHOTS === 'false' &&
+        process.env.UPDATE_ALL_SCREENSHOTS === 'false') ||
+      pass
+    ) {
       this.message = `Screenshots ${
         pass ? 'Matched' : 'Match Failed'
       } for ${fileName} with a tolerance of ${this.expected}%, actual was ${value}%.`
+    } else if (process.env.UPDATE_SCREENSHOTS === 'true') {
+      this.message = 'User was prompted to update screenshot.'
+      pass = true
+    } else if (process.env.UPDATE_ALL_SCREENSHOTS === 'true') {
+      this.message = 'Initial screenshot match failed, but baseline screenshot was updated.'
+      pass = true
     }
 
     return pass
--- a/e2e/assertions/compareScreenshot.js
+++ b/e2e/assertions/compareScreenshot.js
@@ -2,6 +2,8 @@
 const resemble = require('node-resemble-js')
 const fs = require('fs')
 const path = require('path')
+const readline = require('readline')
+const { spawnSync } = require('child_process')
 
 const { tolerance } = require('../config')
 const { getVisualRegressionFolders } = require('../utils')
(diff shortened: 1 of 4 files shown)


e2e/assertions/compareScreenshot.js
```diff
@@ -15,6 +17,26 @@ const ensureBaselinePhotoExists = (baselinePath, resultPath) => {
...
+    output: process.stdout,
+  })
+
+  read.question('Would you like to update the baseline for this screenshot? (y/n) ', answer => {
+    if (answer.toLowerCase() === 'y' || answer.toLowerCase() === 'yes') {
+      console.log('Generating new baseline screenshot...')
+      spawnSync('rm', [baselinePath], {
+        stdio: 'inherit',
+      })
+      fs.writeFileSync(baselinePath, fs.readFileSync(resultPath))
```
ryanoglesby08: Cool! You may not need to manually to the "rm" first though. 

According to the node docs for `writeFile`... https://nodejs.org/api/fs.html#fs_fs_writefile_file_data_options_callback
> Asynchronously writes data to a file, replacing the file if it already exists.

e2e/commands/checkAccessibility.js
```diff
@@ -1,7 +1,7 @@
 const { rootSelector } = require('../config')
 
 exports.command = function checkAccessibility() {
-  this.initAccessibility().assert.accessibility(rootSelector, {
+  this.initAccessibility().verify.accessibility(rootSelector, {
```
ryanoglesby08: What's going on here?

marcod1419: When you use assert, the script will exit on failure. Verify does the exact same thing as assert except that failures don't exit the script.

scripts/e2e.js
```diff
@@ -7,18 +7,15 @@ const getUpdatedPackageNames = require('./utils/getUpdatedPackageNames')
...
         PACKAGES: onlyCorePackages,
+        UPDATE_SCREENSHOTS: process.argv[2] === '-u',
       }),
       stdio: 'inherit',
     }
   )
-
-  if (status !== 0) {
-    process.exit(status)
-  }
```
ryanoglesby08: Please add this stuff with the `status` back. That was there on purpose for the `yarn prepr` task.

If the `spawnSync` for nightwatch here fails, we want the entire task to fail so that the `prepr` task stops, otherwise it ends up returning a 0 exit code, which is interpreted as a success.

e2e/assertions/compareScreenshot.js
marcod1419: Oh! Awesome! :)

scripts/e2e.js
marcod1419: Ah, I see. Got it, I'll uncomment it.

e2e/commands/checkAccessibility.js
ryanoglesby08: 🆗

scripts/e2e.js
```diff
@@ -13,6 +13,8 @@ getUpdatedPackageNames(packageNames => {
     {
       env: Object.assign({}, process.env, {
         PACKAGES: onlyCorePackages,
+        UPDATE_SCREENSHOTS: process.argv[2] === '-i',
+        UPDATE_ALL_SCREENSHOTS: process.argv[2] === '-u',
```
ryanoglesby08: I think this may break down when specifying the components to test. You can either run the e2e tests for all components or individual ones.

```bash
# run on changed components using `lerna updated`
yarn e2e

# run on only Button and Link
yarn test:e2e @tds/core-button @tds/core-link
```

So the `process.argv[2]` would break down in the latter case.

We already have the minimist package for parsing command line arguments though, and it is used in "scripts/utils/parseArgs.js". Check out that and see how we can use that here too.

e2e/assertions/compareScreenshot.js
```diff
@@ -57,11 +88,23 @@ exports.assertion = function(componentName, fileName) {
...
+      (process.env.UPDATE_SCREENSHOTS === 'false' &&
+        process.env.UPDATE_ALL_SCREENSHOTS === 'false') ||
+      pass
+    ) {
+      this.message = `Screenshots ${
+        pass ? 'Matched' : 'Match Failed'
+      } for ${fileName} with a tolerance of ${this.expected}%, actual was ${value}%.`
+    } else if (process.env.UPDATE_SCREENSHOTS === 'true') {
+      this.message = 'User was prompted to update screenshot.'
+      pass = true
```
ryanoglesby08: What if you were prompted and didn't respond with a "yes"?

e2e/assertions/compareScreenshot.js
```diff
@@ -15,6 +16,27 @@ const ensureBaselinePhotoExists = (baselinePath, resultPath) => {
   }
 }
 
+const update = (baselinePath, resultPath, callback, data) => {
+  if (process.env.UPDATE_SCREENSHOTS === 'true') {
```
ryanoglesby08: I feel like it makes sense to assign the `process.env.UPDATE_SCREENSHOTS` globals to `const` variables instead of continually referring to them directly from `process`.

e2e/assertions/compareScreenshot.js
marcod1419: My initial idea for that is since the user actively approved a failing screenshot by choosing not to update it, it would count as a "pass". Although I was probably overthinking what one would consider a pass, so I fixed this in a commit I'm about to push.

e2e/assertions/compareScreenshot.js
```diff
@@ -2,10 +2,16 @@
 const resemble = require('node-resemble-js')
 const fs = require('fs')
 const path = require('path')
+const readline = require('readline')
 
 const { tolerance } = require('../config')
 const { getVisualRegressionFolders } = require('../utils')
 
+const updateScreenshots = process.env.UPDATE_SCREENSHOTS
+const updateAllScreenshots = process.env.UPDATE_ALL_SCREENSHOTS
```
ryanoglesby08: You can simplify further by doing the `=== 'true'` comparison here.

```js
const updateScreenshots = process.env.UPDATE_SCREENSHOTS === 'true'
const updateAllScreenshots = process.env.UPDATE_ALL_SCREENSHOTS === 'true'
```

e2e/assertions/compareScreenshot.js
```diff
@@ -57,11 +94,23 @@ exports.assertion = function(componentName, fileName) {
...
+      this.message = `Screenshots ${
+        pass ? 'Matched' : 'Match Failed'
+      } for ${fileName} with a tolerance of ${this.expected}%, actual was ${value}%.`
+    } else if (updateScreenshots === 'true') {
+      this.message = 'User updated failing baseline screenshot.'
+      pass = true
+    } else if (updateAllScreenshots === 'true') {
+      this.message = 'Initial screenshot match failed, but baseline screenshot was updated.'
+      pass = true
+    }
```
ryanoglesby08: This if...else... block doesn't seem to contain a case for `pass === false`? What happens if screenshots are not being updated and the screenshot failed, or is that not a valid case?

e2e/assertions/compareScreenshot.js
```diff
@@ -43,8 +72,16 @@ exports.assertion = function(componentName, fileName) {
       .compareTo(resultPath)
       // .ignoreAntialiasing()
       // .ignoreColors()
-      .onComplete(callback)
-
+      .onComplete(data => {
+        if (
+          Number(data.misMatchPercentage) > 0.01 &&
```
ryanoglesby08: I just noticed this and I don't understand it. What is `data.misMatchPercentage`? Is that the difference in the comparisons? If so, it should use the `tolerance` variable and not hard coded to 0.01.

e2e/assertions/compareScreenshot.js
marcod1419: Good idea!

ryanoglesby08:
I think this looks good now, I'm just wondering from a workflow point of view how I would use these options?

Let's say I edit the Notification component and remove the Icon, on purpose...

First, I run `yarn test:e2e` and the visual regression screenshot "fails".

Second, I open up the baseline, result, and diff images and look to see what the failure was. Since I  purposefully removed the Icon, I expected this to happen, and now I've confirmed it.

Third, I run `yarn test:e2e -u` or to be more specific I can run `yarn test:e2e @tds/core-notification -u` to update the failing screenshot and see output telling me that the Notification screenshot was updated. At this point I should have a passing e2e test and a local change of a new screenshot.

When would I use `yarn test:e2e -i`?

marcod1419:
@ryanoglesby08 I agree that the usage of `-i` is a bit more of an edge-case, although I do see some value in having it as an option. For example, a situation where you have a very large amount of snapshots that need to be updated. Some snapshots in that large pool may be ok to update, while some are legitimate errors and should not be updated. This command then comes in handy because it means the user will not have to type out a very large list of package screenshots to update with the -u command, and avoid potential user error. In short, I see this option as more of a guided, safer way of vetting a large amount of changes.

ryanoglesby08:
Maybe @theetrain wants to add his thoughts to this discussion?

Ok I see. However, is the `-i` flow that you describe something that we need to solve NOW and that we don't already have a workflow for?

> For example, a situation where you have a very large amount of snapshots that need to be updated. Some snapshots in that large pool may be ok to update, while some are legitimate errors and should not be updated.

Have we run into this yet? I'm wondering if this is in edge-case that we have not yet encountered, in which case I would not want to solve for it yet. The more likely case I can imagine would be a large number of snapshots pass, and 1 fails, in which case the `yarn test:e2e <component name> -u` works great. If we have large numbers of snapshots all failing at the same time, I feel like it means we are doing something wrong...

I usually lean on the side of NOT implementing things that we think we will need later. In this case it may seem innocuous, but means we now have another potential way to run visual regression and to update screenshots. As a dev, I'd prefer to have less options so that I don't have to think about it as much.

My recommendation is to have either `-i` or `-u`, and we use it in coordination with the option of passing component names in on the command line, and NOT have both `-i` and `-u`.

I'll leave it to you @marcod1419 though and approve the PR now.

Issue #535, opened 2018-03-22
[ExpandCollapse] content within collapsed Panels can be read/tab-focused


theetrain:
## Description

Originally reported by @owestinTelus 

While an ExpandCollapse component has a collapsed Panel containing a hyperlink, the link is keyboard-focusable. Also, screen readers may read the contents of collapsed Panels.

## Reproduction Steps

1. Go to https://tds.telus.com/components/index.html#!/ExpandCollapse/0
2. Paste the following code
```jsx
<ExpandCollapse>
  <ExpandCollapse.Panel id="features" header="Features">
    <Box between={3}>
      <Box between={2}>
        <Heading level="h4">Connected GPS</Heading>
        <Paragraph size="medium">
          Connect to your phone's GPS to see real-time run stats. <Link href="telus.com">Go to telus</Link>
        </Paragraph>
      </Box>

      <Box between={2}>
        <Heading level="h4">Notifications</Heading>
        <Paragraph size="medium">
          See call, text and calendar alerts on your wrist.
        </Paragraph>
      </Box>
    </Box>
  </ExpandCollapse.Panel>

  <ExpandCollapse.Panel id="specs" header="Specifications">
    <Box between={3}>
      <Box between={2}>
        <Heading level="h4">Display</Heading>
        <Paragraph size="medium">OLED</Paragraph>
      </Box>

      <Box between={2}>
        <Heading level="h4">Battery Life</Heading>
        <Paragraph size="medium">Up to 5 days</Paragraph>
      </Box>
    </Box>
  </ExpandCollapse.Panel>
</ExpandCollapse>
```
3. While all Panels are collapsed, use the keyboard TAB button
4. Observe the hyperlink becomes keyboard-focusable while its parent Panel is collapsed

## Reproduction Percentage (if applicable)

* 10/10 times

## Meta

* TDS version: all since ExpandCollapse's first release
* Willing to develop solution: Yes
* Has workaround: No
* High impact: No

## Screenshots

![image](https://user-images.githubusercontent.com/12798751/37798935-979a6340-2df4-11e8-9add-1192a230d4d0.png)

ryanoglesby08:
Fixed in v0.34.13, v1.0.12, `@tds/core-expand-collapse/@1.0.2`

PR #586, opened 2018-04-17
fix(core-expand-collapse): prevent tabbing to items while collapsed

marcod1419:
Addresses #535.

After experimenting with a couple of different methods for fixing this, this seems like the cleanest solution I have found so far. Along with fixing items being able to be tabbed into while the component is collapsed, it also allows the `Reveal` animation component to pass in its `status` to children. While we'll have to update a shared component doing this, it could provide some useful functionality down the road.

Changes:
 - `@tds/core-expand-collapse`: 1.0.1 => 1.0.2
 - `@tds/core-input`: 1.0.3 => 1.0.4
 - `@tds/core-select`: 1.0.4 => 1.0.5
 - `@tds/core-text-area`: 1.0.3 => 1.0.4
 - `@tds/shared-animation`: 1.0.0 => 1.0.1 (private)

This can be rebased into master.


ryanoglesby08:
Not sure if you saw this, or if its anything new, but I found this page to be very informative: http://developer-guide.access4all.ch/en/examples/hidden_content

ryanoglesby08:
Seems like the proposed solution is to remove the panel content from the DOM when the panel is closed, and render it when the panel is opened. Seems like a good strategy overall, a couple comments on the implementation though.

Would a simpler solution be to have the Reveal component itself control the visibility of its children? I feel that the intention is that when Reveal is "closed" then the content should be inaccessible. Is an even simpler solution to have the Reveal component set `display: none;` when it is closed in addition to or instead of `height: 0` and `overflow: hidden`? Or, if that doesn't work, to have the Reveal component just not render its children when closed?

ryanoglesby08:
Also, if we are changing some behaviour, we will need a unit test to change other than snapshots. Maybe we need to add one for new behaviour?

marcod1419:
@ryanoglesby08 That was informative! Upon further inspection, this is a viable solution. However, should we implement `aria-hidden` on the animation level? I think it would be a good idea to make that a default behaviour of the reveal animation, so I'm for your suggestion.

marcod1419:
@ryanoglesby08 (Deleted my previous reply, turns out `aria-hidden` seems to actually not work) 

I do agree with making this the default behaviour of `Reveal` though, so I'll see to implementing it that way. (Unless anyone can think of a reason for not hiding elements while `Reveal` is closed?)

ryanoglesby08:
Re: `aria-hidden` -- I think having aria hidden in addition to display none is still a good idea. I think its just that not all browsers or screen readers behave the same way (as is the case for lost of a11y stuff). For example, Chrome works now when all we have is aria-hidden. I think its best practice to have it too.

Yeah, I'm not sure of a reason to not hide elements when its closed. This animation is only used for ExpandCollapse right now, so it will be a low impact change.

marcod1419:
@ryanoglesby08 Moved the behaviour to `Reveal`. Is a test necessary for this? I think a screenshot might be enough since all it does is add a CSS class. Also, I was able to use `visibility` instead of `display`, which is nice for performance. :)

marcod1419:
I think I'll also add the `aria-hidden` logic to `Reveal` and remove it from `ExpandCollapse`, since we're giving `Reveal` the functionality of hiding everything.

ryanoglesby08:
>  I think a screenshot might be enough since all it does is add a CSS class.

I assume you mean snapshot? As in a jest snapshot test? :) If so, then yeah I suppose, seems like adding the inline style and updating the Reveal snapshot works.

And yep, agree with moving the aria-hidden logic to Reveal. Though for that I think that will need some JS logic, in which case I think a dedicated unit test for it would be good.

marcod1419:
@ryanoglesby08 Yeah sorry, that's what I meant. I get those mixed up sometimes haha.

And yep! Working on that now! :)

marcod1419:
Added prepr versioning results and merge preference to the body.

packages/ExpandCollapse/__tests__/ExpandCollapse.spec.jsx
```diff
@@ -249,30 +249,6 @@ describe('ExpandCollapse', () => {
...
-      const { findPanelHeader, clickPanel } = doMount(
-        <ExpandCollapse>{aPanel({ id: 'panel-1' })}</ExpandCollapse>
-      )
-
-      expect(findPanelHeader('panel-1')).toHaveProp('aria-expanded', false)
-
-      clickPanel('panel-1')
-      expect(findPanelHeader('panel-1')).toHaveProp('aria-expanded', true)
-    })
-
```
ryanoglesby08: Why were the `aria-expanded` tests removed? Did you also remove `aria-expanded` from ExpandCollapse? If so, why? If not, please add this test back.

shared/components/Animation/__tests__/Reveal.spec.jsx
```diff
@@ -24,3 +24,25 @@ describe('Reveal', () => {
     expect(reveal).toMatchSnapshot()
   })
 })
+
+describe('accessibility', () => {
+  it('marks the panel as an expandable trigger for content for screen readers', () => {
```
ryanoglesby08: This test description seems to be referring to `aria-expanded`...

Maybe it should read `it('shows panel content to screen readers when opened')`

shared/components/Animation/__tests__/Reveal.spec.jsx
```diff
@@ -24,3 +24,25 @@ describe('Reveal', () => {
     expect(reveal).toMatchSnapshot()
   })
 })
+
+describe('accessibility', () => {
```
ryanoglesby08: Small picky thing... our test convention is to next all tests within a top level `describe`. Nested `describe`s are fine. 

So this test file may look like:

```js
describe('Reveal', () => {
  it('test 1', () => {
  })

  it('test 2', () => {
  })

  describe('nested behaviours only if necessary because there are a lot of tests', () => {
    it('test 3', () => {
    })
  })
})
```

But in this case, there are only a few tests so the nested `describe` is probably overkill.

shared/components/Animation/__tests__/Reveal.spec.jsx
```diff
@@ -24,3 +24,25 @@ describe('Reveal', () => {
...
+  it('marks the panel as an expandable trigger for content for screen readers', () => {
+    const reveal = shallow(
+      <Reveal timeout={500} in={true} height={300}>
+        {() => <div>Content to reveal</div>}
+      </Reveal>
+    )
+
+    expect(reveal.dive()).toHaveProp('aria-hidden', false)
```
ryanoglesby08: Doing `reveal.dive()` is a pattern we used to do a lot, but have starting moving away from because it is pretty brittle and relies on an exact HTML structure, so its likely to break when you don't want it to.

Instead, try placing a `data-testid="some id for your tests"` on the appropriate element in the Reveal component and then use a selector to target it in your test. We also will create a `doShallow` or `doMount` function to encapsulate rendering. This is especially helpful to hide away props that don't matter for the behaviour you are testing. In this case you set `timeout` and `height` because they are required, but those don't matter for this test.

Example: https://github.com/telusdigital/tds-core/blob/master/packages/Tooltip/__tests__/Tooltip.spec.jsx#L16-L27

Let me know if you have questions about this stuff. It is probably easier to show than describe in a comment.

shared/components/Animation/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@tds/shared-animation",
+  "version": "1.0.0",
```
ryanoglesby08: I'm not sure if this should start at `1.0.0` or at `0.1.0`..... Will lerna need to do a release and bump it to version 1?

shared/components/Animation/Reveal.jsx
```diff
@@ -13,9 +13,14 @@ const transitionStyles = height => {
     height: `${height}px`,
   }
 
+  const hiddenContent = {
+    visibility: 'hidden',
+  }
+
```
ryanoglesby08: Let's hoist `hiddenContent` out of the `transitionStyles` function since it never changes

marcod1419:
To quote DJ Khaled, "I've changed. A LOT. Your pull requests can too."

Or something like that. Anyway, I've updated the PR with the test changes and hopefully we're good to go now. :)

shared/components/Animation/__tests__/Reveal.spec.jsx
```diff
@@ -1,26 +1,37 @@
...
-        {() => <div>Content to reveal</div>}
-      </Reveal>
-    )
+    const reveal = doMount({ in: false })
 
     expect(reveal).toMatchSnapshot()
   })
+
+  it('marks the content as visible to screen readers when expanded', () => {
+    const reveal = doMount({ in: true })
```
ryanoglesby08: Nice! This is good. The only prop that we care about here is `in`. 👍

shared/components/Animation/__tests__/Reveal.spec.jsx
```diff
@@ -1,26 +1,37 @@
...
 
 import Reveal from '../Reveal'
 
 describe('Reveal', () => {
-  it('renders when the transition is triggered', () => {
-    const reveal = render(
-      <Reveal timeout={500} in={true} height={300}>
+  const doMount = (overrides = {}) =>
+    mount(
+      <Reveal {...overrides} timeout={500} height={300}>
```
ryanoglesby08: Small thing here. The overrides should be applied AFTER the defaults. Otherwise they are not really overrides. :)

shared/components/Animation/__tests__/Reveal.spec.jsx
```diff
@@ -1,26 +1,37 @@
...
         {() => <div>Content to reveal</div>}
       </Reveal>
     )
 
+  it('renders when the transition is triggered', () => {
+    const reveal = doMount({ in: true })
+
```
ryanoglesby08: Actually, I think we are better off doing a `render` for the snapshot instead of a `mount`. If you look at the snapshot now it is way large and contains the entire component tree, which is less useful for a snapshot like this. Having the entire tree makes it harder to see what's going on.

I recommend just leaving the snapshot tests the way they were before, as we care about all the props for the snapshot.

PR #589, opened 2018-04-19
Test/update baseline images

theetrain:
- update all baseline images due to full-width documentation
- add `-a` flag to test all packages when running `yarn test:e2e`
- fix typo in `@tds/css-reset` import statement


packages/css-reset/CssReset.md
```diff
@@ -1,7 +1,7 @@
 ```js static
 // App.jsx
 
-import '@tds/core-css-reset/dist/index.css'
+import '~@tds/core-css-reset/dist/index.css'
```
ryanoglesby08: Actually, the "~" is not needed here. Please remove.

This import statement is in a jsx file. The "~" is specific to Sass imports.

See starter kit here: https://github.com/telusdigital/telus-isomorphic-starter-kit/blob/master/ui/src/components/App/index.jsx#L4

scripts/e2e.js
```diff
@@ -3,15 +3,16 @@
 // Usage notes
 // Nightwatch CLI options http://nightwatchjs.org/guide#command-line-options
 // Pass -u to update baseline images while running e2e tests
+// Pass -a to test all packages regardless if they've been updated
```
ryanoglesby08: Cool. I see the utility of this, and pretty since solution using `ls` too.

packages/css-reset/CssReset.md
theetrain: I see, I made this in response to another developer's use case who appeared to be adding css reset to their application's main SCSS file. I'll remove the tilde and rely on developer discretion.

PR #598, opened 2018-04-23
chore(dependencies): upgrade rollup and plugins

ryanoglesby08:
* upgrade to v1 of postcss and postcss modules plugins
* realized the `.containsTooltip` class is no longer used, so removed it

postcss-modules still does not properly support Sass files in `composes` statements, so our custom `CssModulesSassLoader` is still needed.

I may try to submit a fix to postcss-modules in the meantime.

### Output of `yarn prepr`

```
$ yarn prepr
yarn run v1.6.0
$ node scripts/e2e.js
lerna info Comparing with @tds/core-expand-collapse@1.0.2.

```


commitlint.config.js
```diff
@@ -13,7 +13,8 @@ const applyCustomScope = () => {
     'guide',
     'tech-snacks',
     'lint',
-    'scaffolding'
+    'scaffolding',
+    'shared-styles'
```
theetrain: How about renaming this to 'shared' to better match the directory name?

package.json
```diff
@@ -102,23 +102,23 @@
...
     "node-sass": "^4.7.2",
-    "node-sass-tilde-importer": "^1.0.1",
+    "node-sass-tilde-importer": "^1.0.2",
     "npm-run-all": "^4.1.2",
     "postcss": "^6.0.17",
-    "postcss-cli": "^4.1.0",
+    "postcss-cli": "^5.0.0",
     "postcss-loader": "^2.1.0",
-    "postcss-modules": "^0.8.0",
+    "postcss-modules": "^1.1.0",
```
theetrain: Is this used anywhere?

commitlint.config.js
theetrain: My thought is to make the scope as generic as possible while referring to a tangible directory name. The commit `chore(shared): remove unused tooltip class` is descriptive enough, and the scope is reusable for different contexts that can be described by the subject. Really though, it's personal preference; the final choice is yours.

The PR is good whether or not you keep this change 👍

ryanoglesby08: Oh, I didn't realize you were trying to base the scopes on the folder names! I think thats a pretty good idea. In that case we are missing some things. Maybe I'll make another PR for that and see what that looks like.

PR #599, opened 2018-04-23
docs(guide): add q3 2018 goals to roadmap

theetrain:
- set q3 roadmap to 'current'
- add q3 roadmap
- move split components item to 'history'
- update headings to use lowercase after first word


theetrain:
We can revisit this.

Issue #582, opened 2018-04-13
[Notification] Links as children do not meet AAA colour contrast compliancy


theetrain:
## Description

When using the Notification component and nesting the TDS Link component, the Link's text colour is a light grey that does not meet AAA WCAG colour contrast standards.

## Reproduction Steps

1. Go to https://tds.telus.com/components/index.html#!/Notification/1
2. Replace the existing example with the following:
```jsx
<Notification variant="branded">
  <Text bold>Tip:</Text> The services are best suited for larger business organizations ordering
  more than 50 plans on one account. <Link href="http://telus.com">Go to TELUS</Link>
</Notification>
```
3. Inspect the Link component and note the colour contrast 

## Reproduction Percentage (if applicable)

* 10/10 times

## Meta

* TDS version: all supported versions
* Willing to develop solution: Yes
* Has workaround: No
* High impact: No

## Screenshots

![image](https://user-images.githubusercontent.com/12798751/38742472-a2850f5e-3f0a-11e8-8da1-ff8d27d62037.png)

![image](https://user-images.githubusercontent.com/12798751/38742423-7e82e52c-3f0a-11e8-9b95-44f44683dbaf.png)

ryanoglesby08:
Per design direction, the fix will be to change the color of all links (not including chevron links) to be the same as body text, "color-shark".

marcod1419:
Fixed in this release: https://github.com/telusdigital/tds-core/releases/tag/%40tds%2Fshared-form-field%401.0.4

PR #601, opened 2018-04-24
fix(core-link): update link colour to be AAA WCAG compliant

marcod1419:
Just updates the Link text colour to be AAA WCAG compliant.

Changes:
 - `@tds/core-link`: 1.0.1 => 1.0.2

This should be squashed.


packages/colours/_colours.scss
```diff
@@ -17,7 +17,7 @@ $color-white: #fff;
 $color-gainsboro: #d8d8d8;
 
 // Typography colors
-$color-link: $color-shuttle-grey;
+$color-link: $color-shark;
```
ryanoglesby08: Hmmmmm. Is this maybe a lazy way of doing this?

Maybe the `$color-link` variable is just no longer needed and can be deleted?

Maybe the Link component should just use the `$color-text` token in "Link.modules.scss"?

Thoughts?

ryanoglesby08:
@marcod1419 Please run `yarn prepr` and paste the output here.

I'm assuming that some visual regression screenshots will need to be updated.

packages/colours/_colours.scss
marcod1419: Yeah, I thought about this. Removing it would mean we would have to update the `Link` component as well. Plus keeping it allows us to easily change the link colour in the future. It doesn't have any impact on the way the styles actually compile, so leaving it seems fairly harmless.

theetrain: I see these variables as a form of documentation since it adds verbosity to otherwise generic colours, just like how we have `$color-icon-secondary` below being referenced by the Icon styles rather than directly referencing `$color-secondary` or even `$color-telus-purple`.

https://github.com/telusdigital/tds-core/blob/48836eb97407ae9796142f05c8ecdb2e97f80883/shared/components/Icon/Icon.modules.scss#L7-L9

This change affects all consuming components anyway, but the only tangible impact is on the developer consuming these colours. Lastly, removing `$color-link` is a breaking change, and this package is being distributed.

ryanoglesby08: Hmmmm, I see this a little differently. I think the semantically named colours being used here are a relic of days past, when TDS was a giant CSS file, and before components, and this is something we should be moving away from. Taking the icon color example, there is one Icon component and all the knowledge of it should be contained therein.

By design, the Icon is in full control of its own styling, including colours, while the colours package defines the core colour palette. Any semantic naming should be done by the components themselves, not from the colours package, which would spread out that knowledge. If a colour has a specific purpose for a specific component, I think that component control it.

This hasn't come up until now, but this seems like a good conversation to have now. :)

theetrain: That all sounds agreeable to me. How about this:

- deprecate the verbose colour variables in `core-colours` using `@warn` (minor release)
- refactor consuming components to use the raw colour variables (patch releases)
- any new release of core-colours can be major

Thoughts?

ryanoglesby08: > deprecate the verbose colour variables in core-colours using @warn (minor release)

Deprecating Sass variables hasn't worked out well for us in the past. Consumers will see the warning every time their app compiles, with no way to turn it off, even if they are not using the variables. People complained ALOT, so we removed all Sass warnings.

Agree with the 2nd two steps.

Might be a bigger change than this story needs. So maybe for this one just focusing on the link is fine? And we can address the other stuff with a follow up story?

marcod1419: Updated the PR with reverting `core-colours` and adding a colour code to the `Link` component. We'll just have to squash this PR to make sure only the `Link` component gets updated. :)

packages/Link/Link.modules.scss
```diff
@@ -30,7 +30,7 @@
 
   &:link,
   &:visited {
-    color: $color-link;
+    color: #2a2c2e;
```
ryanoglesby08: I guess we had some miscommunication here, sorry about that. We should not be hardcoding the color like this and should be using a variable from the colour palatte, just not `color-link`

I think what we want is `color-shark` or `color-text`. Maybe still using `color-text` is fine because it needs to be shared by multiple components, Headings, Link, Text, etc.

marcod1419: Haha, yeah, just realized that could be the case. I'll change it to `color-shark` to make it more resilient to future changes to `core-colour`.

theetrain:
This relates to #582

PR #608, opened 2018-04-26
chore(e2e): set a device scale factor for chrome

ryanoglesby08:
so that screenshots will have the same size regardless of display resolutions.

This should make screenshots come out at 1280 x height instead of 2560 x height as they were before. It should work if using Mac display, or external display.

### yarn prepr output

```bash
$ yarn prepr
yarn run v1.6.0
$ node scripts/e2e.js
lerna info Comparing with @tds/core-input@1.0.5.

✨  Done in 2.61s.
```


config/nightwatch.conf.js
```diff
@@ -36,7 +36,7 @@ module.exports = {
       desiredCapabilities: {
         browserName: 'chrome',
         chromeOptions: {
-          args: ['--no-sandbox', 'window-size=1280,3000'],
+          args: ['no-sandbox', 'window-size=1280,3000'],
```
theetrain: I'm curious why this works without the `--` in front of the argument name.

PR #609, opened 2018-04-27
chore(e2e): add an e2e environment to styleguidist

theetrain:
- update all baseline images
- now spinner will stop spinning when running e2e tests

`yarn prepr` output:

> No components have been changed, nothing to do. Exiting.
>
> 🤔 If this is not what you expected, ensure that your commit messages follow the Conventional Commits specification (https://conventionalcommits.org) and try again.

Please **rebase and merge**. After merging, please:

- [ ] update our internal deployment documentation to include use of new script `yarn dev:e2e`
- [ ] duplicate this fix to tds-community


ryanoglesby08:
Why did you have to re-generate all the screenshots? None of them should have changed right? Except for spinner...

`yarn test:e2e @tds/core-spinner -u`

config/styleguide.config.js
```diff
@@ -324,6 +324,14 @@ module.exports = {
...
       },
     },
+    StyleGuide: {
+      '@global body *':
+        styleguidistEnv === 'e2e'
+          ? {
+              animationPlayState: 'paused !important',
+            }
+          : {},
+    },
```
ryanoglesby08: This seems a bit odd to me to add a global style like this using the StyleGuide component, but I see what you are trying to do.

Perhaps I can recommend an alternative:

Use the `require` config option to dynamically include an extra CSS file with the above styles.

```css
// e2e.css

body * {
  animationPlayState: paused !important;
}
```

```js
// styleguide.config.js

{
  require: [
    path.resolve('packages/css-reset/index.scss'),
    path.resolve('docs/scss/styleguide.scss'),
    styleguideEnv === 'e2e' && path.resolve('docs/scss/e2e.css')
  ]
}
```

I think this is a bit cleaner and clearer.

package.json
```diff
@@ -23,6 +23,7 @@
     "deploy:package": "scripts/publish.sh",
     "commitmsg": "commitlint -e $GIT_PARAMS",
     "dev": "styleguidist server --config config/styleguide.config.js",
+    "dev:e2e": "STYLEGUIDIST_ENV=e2e styleguidist server --config config/styleguide.config.js",
```
ryanoglesby08: This change makes it more necessary for the `yarn e2e` test start and stop the styleguide too, as it seems likely that we will forget this step.

The `yarn e2e` script should:

1. start the docs server with the "e2e" environment
2. wait for it to fully start
2. run tests
3. close down the docs server, whether the tests passed, failed, or the user exited the process.

All of that is out of scope here for sure. Perhaps you can create a story for that?

config/styleguide.config.js
theetrain: When I do it this way, all the baseline images change again. I suspect it has something to do with a race condition of having e2e.css getting injected into the `<head>` causing the CSSOM to reset and cause a flash of unstyled fonts - which changes the kerning of some text in the new baseline images.

How about we put this out of scope along with your next suggestion? That way we can add a longer timeout period before creating screenshots.

ryanoglesby08:
To the merger: see Enrico's checklist at the top for the steps to follow after merging this in.

config/styleguide.config.js
ryanoglesby08: I'm fine with that, though I don't know about the race condition you outline. Either way, using the `theme` or `require`, the styles are being injected into the <head> at runtime. Seems that using the 2 strategies slightly changes the overall page styling somehow though.

PR #617, opened 2018-05-08
chore(scripts): add contributor list generation

marcod1419:
This will automatically generate a contributor list as part of the pre-commit process. 

This should be squashed.


package.json
```diff
@@ -32,7 +36,7 @@
     "lint:scss": "stylelint '{packages,shared}/**/*.scss' --config config/stylelint.config.js",
     "lint:ec": "echint",
     "lint": "npm-run-all --parallel lint:*",
-    "precommit": "lint-staged && yarn test",
+    "precommit": "lint-staged && yarn test && yarn contributors:update",
```
ryanoglesby08: Maybe its better for this to be part of the prepush instead? That way you don't regenerate the contributors list every time you make a small commit.

package.json
```diff
@@ -22,6 +22,10 @@
     "deploy:docs": "node ./scripts/deploy-docs.js",
     "deploy:package": "scripts/publish.sh",
     "commitmsg": "commitlint -e $GIT_PARAMS",
+    "contributors:add": "all-contributors add",
+    "contributors:check": "all-contributors check",
+    "contributors:generate": "all-contributors generate",
```
ryanoglesby08: I see that `yarn contributors:update` is used in the precommit. But what about the other ones? When do you use add, check, and generate? Do we need to run these at some other time? If not, then perhaps we don't need these extra yarn scripts?

scripts/update-contributors.js
```diff
@@ -0,0 +1,37 @@
...
+
+const trimList = peopleList.map(name => {
+  if (name.trim() !== '') {
+    return name.trim()
+  }
+  return null
+})
+
+if (trimList.length > 0) {
+  trimList.map(name => {
```
ryanoglesby08: Use a `.forEach` here instead of `.map`. `.map` should only be used if you are using the return value. `.forEach` is for general iterating. 👍

scripts/update-contributors.js
```diff
@@ -0,0 +1,37 @@
...
+
+const peopleList = people.output[1].slice(people.output[1].indexOf('\n', 25)).split(',')
+
+const trimList = peopleList.map(name => {
+  if (name.trim() !== '') {
+    return name.trim()
+  }
+  return null
+})
+
```
ryanoglesby08: Perhaps this is a bit cleaner.

```js
peopleList
  // trim whitespace from the names
  .map(name => name.trim())
  // filter out empty names
  .filter(name => name.length > 0)
  // filter out bots
  .filter(name => !filteredNames.includes(name))
  .forEach(name => {
    // add the name :)
  })
```

package.json
theetrain: Does `yarn contributors:update` only update the README without making a commit?

marcod1419: @ryanoglesby08 Sure, sounds good to me.

@theetrain Yeah, exactly.

package.json
marcod1419: I put those in there just in case we would want to run things manually for debugging purposes, but you're right, we can probably just remove these and use `npx` or something.

scripts/update-contributors.js
marcod1419: Totally brain farted and forgot about `.forEach` on that! Thanks, I'll replace it.

package.json
```diff
@@ -35,7 +36,7 @@
     "precommit": "lint-staged && yarn test",
     "prepare": "yarn gitbook:install",
     "prepr": "yarn test:e2e && node scripts/prePr.js --conventional-commits --yes",
-    "prepush": "npm-run-all --parallel build test build-docs:*",
+    "prepush": "npm-run-all --parallel build test build-docs:* && yarn contributors:update",
```
ryanoglesby08: I'll approve this as this is minor, but we are using `npm-run-all` here which will run all following scripts in parallel. So this should be:
`"prepush": "npm-run-all --parallel build test build-docs:* contributors:update"`

QA can make this change.

ryanoglesby08:
Actually I thought of something here, if this happens during precommit or prepush, does it modify any files? Do those files then need to be added to the commit?

marcod1419:
@ryanoglesby08 Oh right! Thinking back that’s actually why I put it there in the first place. I can move it back.

ryanoglesby08:
@marcod1419 

Well, I don't know, if you do it precommit the changed file will still need to be `git add`ed. And its likely that we and contributors will forget that.

Maybe the script should make the change and then do the git add?

PR #623, opened 2018-05-10
fix(core-notification): make notification flex grid gutterless

marcod1419:
This fixes an issue where notification text will not line up with the rest of the page's text.

Changes:
 - `@tds/core-notification`: 1.1.0 => 1.1.1


ryanoglesby08:
As per discussion between @theetrain @marcod1419 and I, @marcod1419 will be re-evalutating the solution here as this one does not address the underlying issue of padding property collision on the Column component.

marcod1419:
@theetrain and I have discovered a larger in scope underlying issue to this problem, so I'll close this pull request while we come up with a more robust solution.

Issue #407, opened 2017-08-15
[StepTracker] - Mobile view - "Step x of y" is hardcoded and displayed only in English


amelcharai:
## Description
- On the step tracker for mobile view, copy "Step x of y" is displayed in English for French

## Reproduction Steps
 - https://www.wcstage.telus.com/my-account/payment/preauthorized/account/ye_rZFV2bLYlFedcXxDH5w/debit
 - Login: ssffh4@telusinternal.com / Passw0rd - BAN: 603163163s

## Reproduction Percentage (if applicable)
- 10/10 times

## Version Number of TDS
- [0.19] Include version of TDS this is occuring on

*note attach screenshots whenever possible*

![image](https://user-images.githubusercontent.com/16292379/29332477-27941d0c-81ce-11e7-92d6-e59daa18de1b.png)

ryanoglesby08:
Reopening this until we actually fix the issue.

theetrain:
This has been resolved in `@tds/core-step-tracker@2.0.0` https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-step-tracker%402.0.0

PR #630, opened 2018-05-15
feat(core-step-tracker): component refactor (#407)

marcod1419:
Addresses #407 

~~**This refactor is _extremely_ a work in progress!** I'll end up changing the way some things here work, and there are a few things that still need to be worked out. Tests, for example. Up until now the goal has been to simply get it up and running in the new format with some of the new features.~~

Something to think on in the meantime though, should we try and police what goes into the `current` prop? Right now the prop will go to any number regardless of how many steps there actually are, so it's up to the developer consuming the component to worry about verifying the number in this prop.

**UPDATE:** I would say at this point the PR is at a place where it can be reviewed. However, it is still in progress!

```
Changes:
 - @tds/core-step-tracker: 1.0.0 => 2.0.0
 - @tds/shared-flexbox: 1.0.0 => 1.0.1 (private)
```

This should be rebased!


ryanoglesby08:
> Something to think on in the meantime though, should we try and police what goes into the current prop?

Good point. Seems like an appropriate thing to do, but its up to you if you want to add that in as it wasn't called out in the original story.  If we use PropTypes, then the [airbnb-prop-types](https://github.com/airbnb/prop-types) package has a `range` validator that could work. Also we could do `Math.min` and `Math.max` in the code to ensure that even if you pass too big/small a number it can't go past `steps.length`.

marcod1419:
@ryanoglesby08 I think I'll do it since I feel like it adds a layer of polish to things. That validator would be perfect!

packages/StepTracker/Step/Step.jsx
```diff
@@ -1,30 +1,35 @@
...
+    cls.push(styles.completed)
+  } else if (status === stepIndex) {
+    cls.push(styles.processing)
+  }
+
   /* eslint-disable jsx-a11y/aria-props */
   return (
     <li
-      className={cls}
+      className={cls.join(' ')}
```
ryanoglesby08: We have a shared util called `joinClassNames` that should be used instead. It does basically the same thing, but also removes empties.

packages/StepTracker/Step/Step.jsx
```diff
@@ -1,30 +1,35 @@
...
+
+const Step = ({ label, status, stepNumber, stepIndex }) => {
+  const cls = [styles.step]
+  if (status > stepIndex) {
+    cls.push(styles.completed)
+  } else if (status === stepIndex) {
+    cls.push(styles.processing)
+  }
+
   /* eslint-disable jsx-a11y/aria-props */
```
ryanoglesby08: Why is eslint being disabled? Can you fix the problems instead?

packages/StepTracker/Step/Step.jsx
```diff
@@ -1,30 +1,35 @@
...
  */
-const Step = ({ label, status, stepNumber }) => {
-  const cls = classNames('step-tracker__step', {
-    'step-tracker__step--completed': status === 'completed',
-    'step-tracker__step--processing': status === 'processing',
-    'step-tracker__step--waiting': status === 'waiting',
-  })
+
+const Step = ({ label, status, stepNumber, stepIndex }) => {
+  const cls = [styles.step]
```
ryanoglesby08: A more conventional pattern with CSS modules is to use the `composes` keyword in the CSS file. So the `completed` and `processing` classes should "compose" the `step` class. Then you won't have to do this array manipulation.

```css
.step {
 // styles
}

.completed {
  composes: step;

  // completed styles
}

.processing {
  composes: step;

  // processing styles
}
```

packages/StepTracker/Step/Step.jsx
```diff
@@ -1,30 +1,35 @@
...
-      className={cls}
+      className={cls.join(' ')}
       aria-label={label}
-      aria-current={status === 'processing' ? 'true' : 'false'}
+      aria-current={status === stepIndex ? 'true' : 'false'}
     >
-      <span className="step-tracker__icon" />
-      <span className="step-tracker__label">
+      <span className={styles.icon}>
+        <span style={{ opacity: status > stepIndex ? 1 : 0 }}>
```
ryanoglesby08: I am thinking this opacity is effectively hiding and showing the icon? Is it perhaps better to just not render the Icon instead? If the circle has a set size, then the presence of the Icon shouldn't change that right?

If you do need to do this, perhaps you could use the shared `Display.modules.scss` classes that we have instead. You could use the `invisible` class to set the visibility rather than using the inline styles. That class name may make the intention here more apparent.

packages/StepTracker/Step/Step.jsx
```diff
@@ -34,21 +39,27 @@ const Step = ({ label, status, stepNumber }) => {
 
 Step.propTypes = {
   /**
-   * The label.
+   * The label of this step.
```
ryanoglesby08: Step is not a user facing component right? If so, it shouldn't be included in styleguidist and therefore does not need any prop descriptions. Just the propType definitions is fine, no need for the comments.

packages/StepTracker/Step/Step.jsx
```diff
@@ -34,21 +39,27 @@ const Step = ({ label, status, stepNumber }) => {
...
-  label: PropTypes.string.isRequired,
+  label: PropTypes.string,
   /**
-   * Each step's status is determined by its parent StepTracker component.
+   * The current step the user is on.
    *
    * @ignore
    */
-  status: PropTypes.oneOf(['waiting', 'processing', 'completed']), // eslint-disable-line react/require-default-props
+  status: PropTypes.number, // eslint-disable-line react/require-default-props
```
ryanoglesby08: Please remove the eslint-disable stuff and fix the eslint things.

packages/StepTracker/Step/Step.jsx
marcod1419: Yeah, it's meant to hide it. I did it because there's less chance of performance loss or jank if we're not putting things in and out of the DOM. That class sounds like a good idea though, so I'll use that.

packages/StepTracker/Step/Step.jsx
marcod1419: Missed that. Looks like removing it doesn't actually change something. Maybe a remnant of an old linting config?

packages/StepTracker/Step/Step.jsx
marcod1419: Trying it again, another issue with pulling it from the DOM selectively results in the step label text shifting upward. I also noticed that the `Icon` component needs to have `visibility: inherit` applied to it for visibility to work. Might be worth looking over the shared `Icon` styling at some point in the future. I guess since we've established we want to do things the React way, I'll try to find a workaround or fix for this.

marcod1419: (Figured out a workaround FYI, we should be good now.)

packages/StepTracker/Step/__tests__/Step.spec.jsx
```diff
@@ -4,14 +4,14 @@ import { shallow } from 'enzyme'
 import Step from '../Step'
 
 describe('<Step />', () => {
```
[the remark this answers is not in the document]
marcod1419: Yeah, I guess now that the `Step` component is no longer public-facing, we can just test that through the `StepTracker` component.

packages/StepTracker/StepTracker.modules.scss
```diff
@@ -0,0 +1,112 @@
+@import '~@tds/core-colours/colours';
+@import '~@tds/core-responsive/responsive';
+
+.container {
+  padding: 0;
+  position: relative;
+  display: flex;
+  justify-content: space-between;
+  margin: 1rem -1.5rem;
```
ryanoglesby08: Can we use `Box` instead of hard coded margins?

packages/StepTracker/StepTracker.modules.scss
```diff
@@ -0,0 +1,112 @@
...
+  list-style: none;
+
+  &::before,
+  &::after {
+    position: absolute;
+    top: 17px;
+    width: calc(50% - 1.07rem);
+    content: ' ';
+    border-bottom: 0.1rem solid $color-shuttle-grey;
+  }
```
ryanoglesby08: What is going on here with the before and after pseudo elements? Lots of scary hard coded numbers here.

packages/StepTracker/StepTracker.modules.scss
```diff
@@ -0,0 +1,112 @@
...
+    font-size: 1.2rem;
+    width: 35px;
+    height: 35px;
+    margin-bottom: 1rem;
+  }
+
+  .label {
+    position: relative;
+    color: $color-shark;
+    font-size: 1rem;
```
ryanoglesby08: Let's use the `Text` component instead of setting the font properties here directly.

packages/StepTracker/StepTracker.modules.scss
```diff
@@ -0,0 +1,112 @@
...
+  color: $color-primary;
+
+  @include mq($from: md) {
+    display: none;
+  }
+}
+
+.mobileLabelStepInfo {
+  color: $color-shark;
+  font-size: 1rem;
```
ryanoglesby08: Let's use the `Text` component here too, and then this class probably won't be necessary.

packages/StepTracker/StepTracker.modules.scss
```diff
@@ -0,0 +1,112 @@
...
+    border: 0.1rem solid $color-primary;
+  }
+}
+
+.processing {
+  composes: step;
+  .icon {
+    background-color: $color-primary;
+    text-align: center;
+    border: 0.1rem solid $color-primary;
```
ryanoglesby08: Both `.icon` class definitions look like they are the same. Perhaps it doesn't need to be nested then?

packages/StepTracker/StepTracker.modules.scss
marcod1419: I'll give it a shot.

packages/StepTracker/StepTracker.modules.scss
marcod1419: This controls the lines that appear before and after each step. That calc makes sure that each line is responsive, but also stays short enough to not go through the circles. I played around with trying a better solution earlier but moved on to higher priority things, I can probably revisit this once I'm done with the more important changes.

marcod1419:
Still to come:

- unit tests for other prop types
- give user a warning for having a greater integer value for `current` than number of steps provided

packages/StepTracker/StepTracker.modules.scss
```diff
@@ -0,0 +1,83 @@
...
+
+  &::after {
+    right: 0;
+  }
+
+  .icon {
+    display: inline-block;
+    border: 0.1rem solid $color-shuttle-grey;
+    border-radius: 50%;
+    font-size: 1.2rem;
```
ryanoglesby08: Setting the font-size here seems off. We always use the Typography components to set font size. Perhaps that can be done instead?

packages/StepTracker/StepTracker.modules.scss
```diff
@@ -0,0 +1,83 @@
...
+    width: 35px;
+    height: 35px;
+    margin-bottom: 1rem;
```
ryanoglesby08: Perhaps a Box can be used instead of setting margin bottom? If we need separate to elements that are stacked, then `<Box between={...}>` should help.

packages/StepTracker/StepTracker.modules.scss
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  .label {
+    position: relative;
+    width: 100%;
+    display: none;
+    text-align: center;
+
+    @include mq($from: md) {
+      display: block;
```
ryanoglesby08: Perhaps we should extract some reusable styles/classes for showing and hiding content at breakpoints. We have the same concept in ExpandCollapse now too. See [PanelWrapper.modules.scss](https://github.com/telusdigital/tds-core/blob/master/packages/ExpandCollapse/PanelWrapper/PanelWrapper.modules.scss#L25-L39)

packages/StepTracker/StepTracker.modules.scss
```diff
@@ -0,0 +1,83 @@
...
+    width: 35px;
+    height: 35px;
+    margin-bottom: 1rem;
+  }
+
+  .label {
+    position: relative;
+    width: 100%;
+    display: none;
+    text-align: center;
```
ryanoglesby08: I think you can achieve the same behaviour of centered text with flexbox and the Flexbox component.... then we won't need as many custom styles here.

```jsx
// will need to add the "center" value for `justifyContent`
<Flexbox direction="row" justifyContent="center">
```

packages/StepTracker/__tests__/StepTracker.spec.jsx
```diff
@@ -0,0 +1,52 @@
...
+    steps: ['Login', 'Purchase', 'Checkout'],
+  }
+
+  const doMount = (overrides = {}) => {
+    const container = mount(<StepTracker {...defaultProps} {...overrides} />)
+    const innerSteps = container.find('[data-testid="innerStep"]')
+
+    return {
+      container,
+      innerSteps,
```
ryanoglesby08: Referencing my comment above about `data-testid`, if we use the step index when creating the testid, then this could become...

```js
return {
  container,
  findStep: (index) => container.find(`[data-testid="step-${index}"]`)
}
```

and then our tests would be...

```js
expect(findStep(0)).toContainReact(
  <Icon symbol="checkmark" size={16} variant="inverted" />
)
```

packages/StepTracker/__tests__/StepTracker.spec.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+  it('contains defined steps', () => {
+    const stepTracker = doMount()
+
+    expect(stepTracker.innerSteps.length).toBe(3)
+  })
+
+  it('renders a single step in progress', () => {
+    const { container, innerSteps } = doMount({ current: 2 })
+    expect(container.find('div.stepActive').length).toBe(3)
```
ryanoglesby08: We strongly prefer to use `data-testid` for selectors in tests because it creates un-necessary coupling when class names are used.

The only time we should reference class names in tests is in an assertion: `expect(element).toHaveClassName('stepActive')`

packages/StepTracker/StepTracker.jsx
```diff
@@ -1,95 +1,70 @@
...
+   * String for text to display current step progress on mobile devices.
+   * Use %current to place the current step and use %total to place the total amount of steps.
    */
-  className: PropTypes.string, // eslint-disable-line react/require-default-props
+  stepText: PropTypes.string,
 }
 
 StepTracker.defaultProps = {
-  current: 0,
+  stepText: 'Step %current of %total:',
```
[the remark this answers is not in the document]
marcod1419: I made this choice so people translating it had the option to use the colon or change it to something else if they so desire. It looks less nice but it allows for greater flexibility.

packages/StepTracker/StepTracker.jsx
```diff
@@ -1,95 +1,70 @@
...
   /**
-   * One or more CSS class names separated by spaces to append onto the container.
-   * Don't advertise as we plan on removing this feature soon.
-   *
-   * @ignore
+   * String for text to display current step progress on mobile devices.
+   * Use %current to place the current step and use %total to place the total amount of steps.
    */
-  className: PropTypes.string, // eslint-disable-line react/require-default-props
+  stepText: PropTypes.string,
```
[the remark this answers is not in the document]
marcod1419: What about `mobileStepLabelTemplate`? I want to make it clear on first glance what exactly this label is for.

packages/StepTracker/StepTracker.jsx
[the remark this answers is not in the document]
marcod1419: I like this!

packages/StepTracker/StepTracker.modules.scss
marcod1419: Sounds like a good idea. Since it impacts that component as well, should that be part of this pull request? I feel like it could be its own thing.

packages/StepTracker/StepTracker.modules.scss
marcod1419: Flexbox only allows a `justifyContent` value of `spaceBetween` though, which doesn't seem like something we want.

packages/StepTracker/StepTracker.jsx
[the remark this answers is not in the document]
ryanoglesby08: But what is the colon for? Is there more to the string template? Does something come AFTER the total? Maybe the full template should be: `Step ${current} of ${total}: ${stepLabel}`?

packages/StepTracker/StepTracker.modules.scss
ryanoglesby08: Up to you how you would want to manage that.

packages/StepTracker/StepTracker.modules.scss
ryanoglesby08: We are in full control over the Flexbox component! :) And it is NOT a public component so there is no hesitation to changing the interface. Is there any reason not to add a new accepted value to the `justifyContent` prop?

packages/StepTracker/StepTracker.jsx
```diff
@@ -1,95 +1,70 @@
...
-        </ul>
-        <div className="step-tracker__mobile-label">
-          <span className="step-tracker__mobile-label-step-info">
-            Step {currentStepNumber} of {totalSteps}:
-          </span>{' '}
-          {currentStepLabel}
-        </div>
-      </div>
-    )
+  const getStepLabel = () => {
```
ryanoglesby08: Let's hoist these functions outside of the component definition. The convention is to put all functions at the same level in the file (aka module). This also makes it more clear that these functions have specific input and output.

packages/StepTracker/StepTracker.jsx
```diff
@@ -1,95 +1,70 @@
...
-class StepTracker extends Component {
-  getStatus(current, index) {
-    if (index < current) {
-      return 'completed'
-    } else if (index === current) {
-      return 'processing'
-    }
 
-    return 'waiting'
+const StepTracker = ({ current, steps, mobileStepLabelTemplate }) => {
```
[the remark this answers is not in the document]
ryanoglesby08: We also have 2 common tests that we use for every component to ensure this behaviour is there. You can copy those tests from another component and then implement it. [Here is an example in Button](https://github.com/telusdigital/tds-core/blob/master/packages/Button/__tests__/Button.spec.jsx#L50-L62)

packages/StepTracker/StepTracker.jsx
```diff
@@ -1,95 +1,70 @@
-import React, { Component } from 'react'
+import React from 'react'
 import PropTypes from 'prop-types'
-import classNames from 'classnames'
+
+import Text from '@tds/core-text'
+import Flexbox from '../../shared/components/Flexbox/Flexbox'
```
ryanoglesby08: Should we be marking Flexbox as a private package now and declaring it as a dev dependency? I think thats what we have been doing as we are touching the shared packages.

ryanoglesby08: Ahhhh yes, you are correct. I thought you were adding a prop value to the Flexbox component so that we could do `<Flexbox justifyContent="center">`

packages/StepTracker/Step/Step.jsx
```diff
@@ -1,54 +1,47 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import classNames from 'classnames'
+
+import Icon from '@tds/core-decorative-icon'
+import Text from '@tds/core-text'
+import Flexbox from '../../../shared/components/Flexbox/Flexbox'
```
marcod1419: (Still need to update this with a package import)

packages/StepTracker/StepTracker.md
```diff
@@ -9,21 +9,28 @@ window to see this behavior.
 
 ```jsx
 initialState = {
-  current: 1,
+  current: 0,
 }
 ;<div>
```
theetrain: Is this semicolon placement correct?

packages/StepTracker/StepTracker.modules.scss
theetrain: This will be addressed in a later story: https://telusdigital.atlassian.net/browse/TDS-633

PR #631, opened 2018-05-15
[core-flex-grid] change prop behaviour

simpleimpulse:
https://telusdigital.atlassian.net/browse/TDS-588

- Made limitWidth true by default
- Removed the centre prop 
- Updated docs to not use limitWidth
- Updated core-notification to reflect core-flex-grid changes

`yarn prepr` output

```
Changes:
 - @tds/core-flex-grid: 1.2.1 => 2.0.0
 - @tds/core-notification: 1.1.1 => 1.1.2
```

Before releasing, update starter-kit to demonstrate a full bleed flex-grid example.

Please rebase when ready to QA.


packages/FlexGrid/FlexGrid.jsx
```diff
@@ -17,16 +16,8 @@ import styles from './FlexGrid.modules.scss'
...
-    deprecate(
-      'core-flex-grid',
-      'The centre prop is deprecated due to the limitWidth prop centring the grid on its own. Please remove the centre prop from your grid definition.'
-    )
-  }
-
-  const getClasses = () =>
-    joinClassNames(styles.flexGrid, centre && styles.centre, limitWidth && styles.limitWidth)
+const FlexGrid = ({ limitWidth, gutter, children, ...rest }) => {
+  const getClasses = () => joinClassNames(styles.flexGrid, limitWidth && styles.limitWidth)
```
ryanoglesby08: Small thing. The `getClasses` function is not needed.... Just call `joinClassNames` directly. No need for the extra indirection.

packages/FlexGrid/FlexGrid.modules.scss
```diff
@@ -2,9 +2,6 @@
 
 .flexGrid {
   padding: 0;
-}
-
-.centre {
   margin: 0 auto;
```
ryanoglesby08: Why is `margin: 0 auto` on both the `flexGrid` AND the `limitWidth` classes? It only needs to be there once right?

packages/FlexGrid/__tests__/FlexGrid.spec.jsx
```diff
@@ -1,7 +1,6 @@
 import React from 'react'
 import { mount, render } from 'enzyme'
 import { Grid } from 'react-flexbox-grid'
-import { deprecate } from '../../../shared/utils/warn'
```
ryanoglesby08: You can also remove the `jest.mock` on line 9 I believe.

packages/FlexGrid/__tests__/FlexGrid.spec.jsx
```diff
@@ -65,31 +64,22 @@ describe('FlexGrid', () => {
...
-    expect(flexGrid).toHaveClassName('centre')
-  })
-
-  it('will show a centre is deprecated warning', () => {
-    doMount({ centre: true })
-
-    expect(deprecate).toHaveBeenCalled()
-  })
-
   it('does not allow custom CSS', () => {
```
ryanoglesby08: I think we should have a test for the case when `limitWidth={false}`. That code path is not unit tested currently.

simpleimpulse:
@ryanoglesby08 Thanks for the feedback, I've made changes accordingly.

Issue #605, opened 2018-04-25
[Headings] Separate heading design from semantics


jackreeves:
## Description

Separate design from semantics so that content owners (non-devs) can adjust the semantic heading levels `<h1 - h6>` in site builder without impacting the design

## Recommendation

Separation of styling from element so could pass through props for heading and styling separatly.

## Meta

* TDS version: v1
* Willing to develop solution: No
* Has workaround: No
* High impact: No

PR #635, opened 2018-05-18
feat(core-heading): decouple semantic heading level from visual size

simpleimpulse:
Add a `size` property to determine visual size. If `size` is unset, assume size is the same as `level`.
Addresses issue: #605

Changes
 - `@tds/core-heading`: 1.0.2 => 1.1.0


packages/Heading/Heading.jsx
```diff
@@ -30,17 +30,16 @@ BaseHeading.propTypes = {
  *
  * @version ./package.json
  */
-const Heading = ({ level, invert, children, ...rest }) => {
+const Heading = ({ level, size = level, invert, children, ...rest }) => {
```
ryanoglesby08: We tend to not use default assignment when defining React components in favor of using defaultProps so that the default value gets automatically documented by react-styleguidist.

In this case I see why that's not possible though since the default value is dynamic.

ryanoglesby08:
@theetrain @simpleimpulse 

Let's be extra thorough and thoughtful when changing the interface to components (adding, removing, changing props). In this case we have added a `size` prop in addition to the `level` prop. `<Heading size="h2" level="h3">` I'm not sure if that is intuitive to TDS users? I think they would have to read and re-read the docs now since "size" and "level" are very similar words.

I recommend we carefully consider the interface here so that it is simple, intuitive, and consistent with other components. Put ourselves in our users' shoes.

Here is some brainstorming on other potential interfaces. (Out of these I like `tag` the best)

```jsx
// Simplest interface
<Heading level="h4">A level 4 heading</Heading>

// `tag` is used by other TDS components such as Box
<Heading level="h4" tag="h2">A level 2 HTML heading with the h4 style</Heading> 

// `overrideTag` is explicit that you are overriding the HTML tag
<Heading level="h4" overrideTag="h2">A level 2 HTML heading with the h4 style</Heading>

// This would be a breaking change...
// `level` matches HTML semantic language for heading tags
// `variant` is consistent with other TDS components such as Notification
<Heading variant="h4" level="h2">A level 2 HTML heading with the h4 style</Heading>
```

theetrain:
When we return to this, we will be using the `level` and `tag` props as proposed above:

```jsx
<Heading level="h4" tag="h2">A level 2 HTML heading with the h4 style</Heading>
```

simpleimpulse:
@theetrain I've revised the prop interface according to recommendations. Please re-review.

packages/Heading/Heading.jsx
```diff
@@ -8,60 +8,40 @@ import HeadingSup from './HeadingSup/HeadingSup'
...
  * Page headings. Renders an HTML `<h1-h4>` element.
  *
  * @version ./package.json
  */
-const Heading = ({ level, invert, children, ...rest }) => {
-  const baseHeadingProps = {
-    ...rest,
-    level,
+const Heading = ({ level, tag = level, invert, children, ...rest }) => {
+  const getClassNames = () => {
```
theetrain: Let's prefer to extracting utility methods from the functional component, similar to Button: https://github.com/telus/tds-core/blob/7485396b857a2c2c2baa657abacacfd9123bba5d/packages/Button/Button.jsx#L9-L15

packages/Heading/Heading.jsx
```diff
@@ -8,60 +8,40 @@ import HeadingSup from './HeadingSup/HeadingSup'
...
+      className: joinClassNames(styles[level], getClassNames()),
+    },
+    children
   )
 }
 
 Heading.propTypes = {
   /**
-   * The heading level.
+   * The stylistic level of the heading.
```
theetrain: Try

> The visual level of the heading. If `tag` is not specified, then `level` determines what HTML element to render.

packages/Heading/Heading.jsx
```diff
@@ -8,60 +8,40 @@ import HeadingSup from './HeadingSup/HeadingSup'
...
    */
   level: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4']).isRequired,
+  /**
+   * The semantic level of the heading.
```
theetrain: Try

> The semantic level of the heading. Renders the specified HTML element, otherwise it matches `level`.

packages/Heading/Heading.md
```diff
@@ -13,5 +19,6 @@
 * Organise headers in a nested order: h1, then h2, then h3
```
theetrain: TELUS standard for Canadian English is `Organize`

packages/Heading/Heading.md
```diff
@@ -13,5 +19,6 @@
...
 * Try **not** to skip levels where possible: h1, then h3
 * Use one h1 per page
+* The semantic html tag and stylistic level of the heading can be decoupled by explicitly setting the `tag` property.
```
theetrain: Try

> The semantic HTML tag and visual appearance of the Heading component can be controlled using the `level` and `tag` props together

PR #638, opened 2018-05-21
[Experimental][Feedback requested] Use jest for visual regression

ryanoglesby08:
**TL;DR: This PR is a sample of a new way of doing visual regression testing that does not need the style guide. It is a WIP, and needs your feedback.**

## Overview

IMHO this is a more comprehensive, more flexible, faster, and more maintainable way of doing visual regression testing in comparison to nightwatch and resemble-js. It uses jest as the test-runner and 2 libraries for doing the visual regression:
* [component-image](https://github.com/corygibbons/component-image) generates an image of a React component
* [jest-image-snapshot](https://github.com/americanexpress/jest-image-snapshot) is a jest matcher for doing image comparisons. It behaves just like regular jest snapshots.

### Details of this PR

This PR implements the strategy for only `Button` as an example. It includes visual regression for all 3 button types, and mobile styles (something not done with our current set up). You do NOT need to run the style guide to run these tests.

## Usage

Visual regression tests could easily get run with the other unit tests, or separated into their own jest test suite. Because they get run with jest, all the normal jest commands apply, such as `--watch` or `-u` to update snapshots or screen shots.

This command will run the test added with this PR.
```
npx jest packages/Button/__tests__/Button.visual.spec.jsx --config config/jest-visual.config.js
```

### Output - tests pass

screen shot 2018-05-21 at 11 46 09 am

### Output - tests fail

screen shot 2018-05-21 at 11 47 57 am

## Advantages:

* Visual regression is now decoupled from the style guide. You can make docs changes without affecting visual regression.
* No need to have the style guide running to do screen shot tests. (Good for running in a CI pipeline)
* Full control over the test setup, which makes it easy to test component variations (such as ExpandCollapse or Tooltip in their open and closed states) or responsive styles.
* Much less code to maintain when compared to our nightwatch implementation.
* Probably easier to to visual regression on non-component packages such as responsive media queries since any component can be rendered to an image.
* Tests and screenshots are co-located with the component and unit tests.

## Disadvantages:

* The component-image library is specific to Google Chrome. We are not doing cross-browser screenshotting anyway though.
* The component-image library does not have many downloads. It is very small though and not a critical path dependency, so forking it wouldn't be too bad if necessary.


theetrain:
## Things I like

- having visual test code co-located with components
- the potential to test visual components quickly and often
- ⭐️ having more consistent test results without being adversely affected by changes in CSS from the docs

## Things that leave me uncertain

- Where do accessibility tests go?
- Can this be defined in a centralized manner? Otherwise it seems to have net-more-code

## A nice-to-have dream

- 🕶 defining a schema or detecting prop variants automatically and having them tested
- cross-browser testing in CI via Saucelabs or similar
- being able to test different states (i.e. hover, expanded, etc)

---

Overall, I think we ought to align with the Digital Platform and see whether Jest can be useful for other projects as well, particularly tds-community. To me, whichever option can achieve the more valuable tests should be chosen.

ryanoglesby08:
@theetrain thanks for the insights! Some replies...

> Things that leave me uncertain... Where do accessibility tests go?

Those would remain in the "e2e" suite for now. Though there are jest-based tools for that as well that seem promising, such as https://github.com/nickcolley/jest-axe.

> Things that leave me uncertain... Can this be defined in a centralized manner? Otherwise it seems to have net-more-code

95% of the code in this approach would be specific tests for the components. Each component should have its own nuanced tests to do visual regression that is specific to that component. For example, Tooltip could test for when the tooltip is open and when its closed. Input could have a test to show the various states (normal, focused, success, error). This is normal testing practice, and I see this as being a GOOD thing as it means more thought gets put into what to test for each component.

The big win here is that all of the hand-written code in "e2e/assertions/compareScreenshot.js", "e2e/commands/compareScreenshot.js" and "e2e/tests/Components.spec.js" doing folder creation, screenshotting, CSS selector targetting, browser management, screenshot comparisoin, etc. goes away and the only custom code we write is the tests that are specific to our components.

> A nice-to-have dream... being able to test different states (i.e. hover, expanded, etc)

This isn't a dream! Haha. With this PR this is now totally possible depending on the component and how those states are implemented! For example, we can do a visual regression for an expanded `ExpandCollapse` since that is controlled via props. Visual regression for a hover state is still hard if its CSS-based, but easy if it is JS/props based.


> Overall, I think we ought to align with the Digital Platform and see whether Jest can be useful for other projects as well, particularly tds-community.

Agreed. It may be useful to run this as a bit of an experiment at first to gather data and better determine usefulness before going all in.

jraff:
@simpleimpulse Please create a JIRA ticket to follow this.

marcod1419:
While we're still looking into this and have this backlogged, once we do this we will most likely start the same concept from scratch. We will however using the findings in this PR as a base.

PR #640, opened 2018-05-21
chore(config): extract and optimize css bundle

ryanoglesby08:
* adds the `mini-css-extract-plugin` and `optimize-css-assets-webpack-plugin` webpack plugins
* `optimize-css-assets-webpack-plugin` uses cssnano to do css minimization
* preserves inline style loading in development so that hot module reload continues to work as expected

Note: `mini-css-extract-plugin` is recommended instead of `extract-text-webpack-plugin` for css extraction. See note here https://github.com/webpack-contrib/extract-text-webpack-plugin#usage


githubjosh:
Any minification stats? Existing file size versus new file size. 

Curious how big a difference this makes. All things being equal, smaller files are a welcome performance improvement.

ryanoglesby08:
@githubjosh 

For clarity, this PR only effects the component catalogue (https://tds.telus.com/components/index.html), which has low traffic and importance to the telus.com property. It does not effect the distributed CSS bundles of the components that applications use.

Nevertheless, good questions. We did not extract a CSS bundle at all before, all styles were inlined. I expect the difference to be fairly small, but we should definitely get some kind of baseline before merging this in. Thanks for the reminder!

jraff:
@simpleimpulse Please create a JIRA ticket to follow this.

Issue #654, opened 2018-06-13
"role=alert" not defined on input field error condition


cglynne:
## Description

* TDS feedback error component does not have `role="alert"` defined. When an error condition occurs on an [input field](https://tds.telus.com/components/index.html#input), this attribute notifies screen readers to read the alert. Without it, the alert does not trigger on a form field validation error to the user and the message is not read aloud.

## Reproduction Steps

* Tested using JAWS screen reader on Windows
* Validated that this parameter actually functions by testing the TD.com (bank) login page. Field level alerts are triggered and read aloud by JAWS when they appear.

## Reproduction Percentage (if applicable)

* 100% 

## Meta

* TDS version: 1.0.5
* Willing to develop solution: Not able to contribute at this time
* Has workaround: No
* High impact: Impacts accessability

## Screenshots

* Include any relevant screenshots

theetrain:
This has been amended in [@tds/core-input-feedback@1.0.2](https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-input-feedback%401.0.2)

PR #673, opened 2018-07-04
fix(core-input-feedback): add role alert to error message

marcod1419:
Adds `role="alert"` to all input errors by default. Addresses #654.

prepr:
```
Changes:
 - @tds/core-checkbox: 1.0.4 => 1.0.5
 - @tds/core-input: 1.0.6 => 1.0.7
 - @tds/core-input-feedback: 1.0.1 => 1.0.2
 - @tds/core-radio: 1.0.4 => 1.0.5
 - @tds/core-select: 1.0.7 => 1.0.8
 - @tds/core-text-area: 1.0.6 => 1.0.7
 - @tds/core-tooltip: 1.0.4 => 1.1.0
 - @tds/shared-choice: 1.0.4 => 1.0.5 (private)
 - @tds/shared-form-field: 1.0.5 => 1.0.6 (private)
```

This can be rebased.

### DIFF
--- a/packages/InputFeedback/InputFeedback.jsx
+++ b/packages/InputFeedback/InputFeedback.jsx
@@ -17,6 +17,7 @@ const InputFeedback = ({ feedback, children, ...rest }) => (
     {...safeRest(rest)}
     inset={3}
     dangerouslyAddClassName={feedback ? styles[feedback] : styles.default}
+    role={feedback === 'error' ? 'alert' : null}
   >
     {children}
   </Box>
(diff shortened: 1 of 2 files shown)


packages/InputFeedback/InputFeedback.jsx
```diff
@@ -17,6 +17,7 @@ const InputFeedback = ({ feedback, children, ...rest }) => (
     {...safeRest(rest)}
     inset={3}
     dangerouslyAddClassName={feedback ? styles[feedback] : styles.default}
+    role={feedback === 'error' ? 'alert' : null}
```
theetrain: Considering this will override any consumer's custom usage of `role`, I don't consider this a breaking change since we originally intended to control the `role` attribute to match the component's state anyway.

PR #681, opened 2018-07-12
feat(core-flex-grid): add reversible grid props

marcod1419:
Allows the parent `FlexGrid` component, along with its child `Row` component to be responsively reversible.

prepr:
```
Changes:
 - @tds/core-checkbox: 1.0.4 => 1.0.5
 - @tds/core-flex-grid: 2.0.2 => 2.1.0
 - @tds/core-input: 1.0.6 => 1.0.7
 - @tds/core-input-feedback: 1.0.1 => 1.0.2
 - @tds/core-notification: 1.1.5 => 1.2.0
 - @tds/core-radio: 1.0.4 => 1.0.5
 - @tds/core-select: 1.0.7 => 1.0.8
 - @tds/core-text-area: 1.0.6 => 1.0.7
 - @tds/shared-choice: 1.0.4 => 1.0.5 (private)
 - @tds/shared-form-field: 1.0.5 => 1.0.6 (private)
```

This can be squashed.


packages/FlexGrid/FlexGrid.jsx
```diff
@@ -16,13 +16,48 @@ import styles from './FlexGrid.modules.scss'
...
+  gutter,
+  xsReverse,
+  smReverse,
+  mdReverse,
+  lgReverse,
+  xlReverse,
+  children,
+  ...rest
+}) => {
+  const calculateReverseLevel = () => {
```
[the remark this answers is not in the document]
marcod1419: Yeah that can work.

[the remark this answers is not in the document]
jraff: Everything else looks good!

packages/FlexGrid/Row/Row.jsx
```diff
@@ -51,13 +73,38 @@ Row.propTypes = {
    * Distribute empty space around columns.
    */
   distribute: PropTypes.oneOf(['around', 'between']),
+  /**
+   * Choose if the item order should be reversed within the 'xs' breakpoint range.
```
theetrain: `true` and `false` values have different effects here. It may be worth explaining them more:

>Choose whether item order should be reversed _from_ the `xs` breakpoint.

packages/FlexGrid/Row/Row.jsx
```diff
@@ -51,13 +73,38 @@ Row.propTypes = {
...
+   */
+  xsReverse: PropTypes.bool,
+  /**
+   * Choose if the item order should be reversed within the 'sm' breakpoint range.
```
theetrain: Try

>Choose whether item order should be reversed _from_ the `sm` breakpoint. When you pass in `false`, the order will be normal from the `sm` breakpoint. By default, it inherits the behaviour set by the preceding prop.

packages/FlexGrid/Row/Row.jsx
```diff
@@ -51,13 +73,38 @@ Row.propTypes = {
...
+   */
+  smReverse: PropTypes.bool,
+  /**
+   * Choose if the item order should be reversed within the 'md' breakpoint range.
```
theetrain: Repeat the above comment here.

packages/FlexGrid/Row/Row.jsx
```diff
@@ -51,13 +73,38 @@ Row.propTypes = {
...
+   */
+  mdReverse: PropTypes.bool,
+  /**
+   * Choose if the item order should be reversed within the 'lg' breakpoint range.
```
theetrain: Repeat the above comment here.

packages/FlexGrid/Row/Row.jsx
```diff
@@ -51,13 +73,38 @@ Row.propTypes = {
...
+   */
+  lgReverse: PropTypes.bool,
+  /**
+   * Choose if the item order should be reversed within the 'xl' breakpoint range.
```
theetrain: Repeat the above comment here.

packages/FlexGrid/Row/Row.md
```diff
@@ -107,3 +107,25 @@ Use flexbox distribution properties to control the negative space around columns
   </FlexGrid.Row>
 </FlexGrid>
 ```
+
+### Reversing item order
+
+On certain layouts, the order of items within the row may need to be reversed. To do this, set the corresponding prop for the target breakpoint range as `true`. If a breakpoint is defined as `true`, it will affect all other breakpoints after it until a latter breakpoint assigned as `false` is found.
+
+In the following example, the items are set to reverse until the `md` breakpoint. Then, once the `xl` breakpoint is expanded to, the items will reverse again.
```
theetrain: > ...Then, at the `xl` breakpoint`, the items will reverse again.

packages/FlexGrid/FlexGrid.jsx
```diff
@@ -39,6 +61,26 @@ FlexGrid.propTypes = {
    * Whether or not to include gutters in between columns.
    */
   gutter: PropTypes.bool,
+  /**
+   * Choose if the item order should be reversed within the 'xs' breakpoint range.
```
theetrain: See comments in row.jsx.

packages/FlexGrid/Row/Row.jsx
```diff
@@ -51,13 +73,38 @@ Row.propTypes = {
    * Distribute empty space around columns.
    */
   distribute: PropTypes.oneOf(['around', 'between']),
+  /**
+   * Choose if the item order should be reversed from the 'xs' breakpoint. When you pass in false, the order will be normal from the xs breakpoint. By default, it inherits the behaviour set by the preceding prop.
+   */
+  xsReverse: PropTypes.bool,
+  /**
+   * Choose if the item order should be reversed from the 'sm' breakpoint. When you pass in false, the order will be normal from the sm breakpoint. By default, it inherits the behaviour set by the preceding prop.
```
theetrain: Not a huge deal to me, but should we remove the word `inherits` and just say
> By default, it does not interfere with settings made to the preceding prop.

packages/FlexGrid/FlexGrid.jsx
```diff
@@ -39,6 +61,26 @@ FlexGrid.propTypes = {
    * Whether or not to include gutters in between columns.
    */
   gutter: PropTypes.bool,
+  /**
+   * Choose if the item order should be reversed from the 'xs' breakpoint. When you pass in false, the order will be normal from the xs breakpoint. By default, it inherits the behaviour set by the preceding prop.
```
theetrain: See comments in Row.jsx

Issue #682, opened 2018-07-13
[Checkbox+Radio] Does not allow numbers as value


jraff:
## Description

* Checkbox and Radio do not render when passing a number into the value prop.
* Number should be supported as it's listed as a valid PropType.

## Reproduction Steps

```
<Radio
    label="Paper bill"
    name="monthly-bill"
    value={1}
/>
```
Error:
```
generateId.js:7 Uncaught TypeError: text.toLowerCase is not a function
    at sanitize (generateId.js:7)
    at Object.postfix (generateId.js:24)
    at getId (Choice.jsx:54)
    at Choice.render (Choice.jsx:192)
    at finishClassComponent (react-dom.development.js:7882)
    at updateClassComponent (react-dom.development.js:7859)
    at beginWork (react-dom.development.js:8233)
    at performUnitOfWork (react-dom.development.js:10215)
    at workLoop (react-dom.development.js:10279)
    at HTMLUnknownElement.callCallback (react-dom.development.js:540)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:579)
    at invokeGuardedCallback (react-dom.development.js:436)
    at renderRoot (react-dom.development.js:10357)
    at performWorkOnRoot (react-dom.development.js:10963)
    at performWork (react-dom.development.js:10916)
    at requestWork (react-dom.development.js:10832)
    at scheduleWorkImpl (react-dom.development.js:10715)
    at scheduleWork (react-dom.development.js:10677)
    at scheduleTopLevelUpdate (react-dom.development.js:11140)
    at Object.updateContainer (react-dom.development.js:11178)
    at eval (react-dom.development.js:15190)
    at Object.unbatchedUpdates (react-dom.development.js:11049)
    at renderSubtreeIntoContainer (react-dom.development.js:15189)
    at Object.render (react-dom.development.js:15254)
    at eval (Preview.js:193)
```

## Meta

* TDS version: @tds/core-radio@1.0.5 and @tds/core-checkbox@1.0.5
* Willing to develop solution: Yes
* Has workaround: Yes
* High impact: Yes

PR #683, opened 2018-07-13
fix(shared-choice): allow value to be a number on Radio and Checkbox

jraff:
## Related issues

See #682 

## Description

Update Radio and Checkbox to allow numbers to be passed as value./

## Package changes

```
$ node scripts/e2e.js
lerna info ignore [ '*.md', '*.spec.jsx', '*.snap', '@tds/core-selector-counter' ]
lerna info Comparing with @tds/core-expand-collapse@1.1.0.

```

### DIFF
--- a/packages/Checkbox/__tests__/Checkbox.spec.jsx
+++ b/packages/Checkbox/__tests__/Checkbox.spec.jsx
@@ -39,6 +39,12 @@ describe('Checkbox', () => {
     expect(checkbox).toMatchSnapshot()
   })
 
+  it('allows numbers as value', () => {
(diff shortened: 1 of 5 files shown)


marcod1419:
Seems good, but it looks like lerna isn't picking up the update properly. I think `generateId` may need to be made into a private package so lerna knows to update everything that uses it. Thoughts @theetrain and @simpleimpulse?

theetrain:
I agree, this should be a private package. Should we combine all utils and create a `@tds/shared-utils` package with named exports, or individual packages such as `@tds/util-generate-id`?

I'd say the former even though it may invoke many deploys in the future, plus we can tree shake them using Rollup.

Issue #688, opened 2018-07-17
[Input] Text is cut off at normal zoom level


theetrain:
## Description

The Input component displaying hanging characters such as `g` get cut off at the bottom (see screenshot).

## Reproduction Steps

1. Visit the Input documentation at 100% zoom level
2. Enter a lowercase `g`
3. Observe the character being cut off at the bottom

## Reproduction Percentage (if applicable)

* 10/10 times

## Meta

* TDS version: core-input@1.0.8
* Willing to develop solution: Yes
* Has workaround: No
* High impact: No

## Screenshots

![image](https://user-images.githubusercontent.com/12798751/42825057-3014c3d4-89af-11e8-969c-b5e8706a009d.png)

jraff:
Fixed in [@tds/core-input@1.0.10](https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-input%401.0.10)

PR #699, opened 2018-07-25
fix(core-input): prevent text cutoff for hanging characters

jraff:
## Related issues

See #688 

## Description

Prevent hanging characters like g, y, p, q, from being having their bottoms cutoff in `<Input>`

## Package changes

```
lerna info ignore [ '*.md', '*.spec.jsx', '*.snap', '@tds/core-selector-counter' ]
lerna info Comparing with @tds/core-select@1.0.11.

Changes:
 - @tds/core-input: 1.0.9 => 1.0.10

lerna info auto-confirmed
```

This should be squashed and merged.


packages/Input/Input.modules.scss
```diff
@@ -11,3 +11,9 @@
   composes: absolute centerVertically from '../../shared/styles/Position.modules.scss';
   composes: iconRight from '../../shared/components/FormField/FormField.modules.scss';
 }
+
+.input {
+  min-height: 3.25rem;
+  max-height: 3.25rem;
+  height: auto;
```
simpleimpulse: Is there a specific reason why the `height` prop is set? Because setting `min-height` and `max-height` should pretty much be sufficient.

jraff: The issue is specifically with `height`. In order to get around it we need to clear the existing height with `height: auto` and instead use `min-height` and `max-height`.

packages/Input/Input.jsx
```diff
@@ -23,7 +23,12 @@ const Input = props => (
       <div className={positionStyles.relative}>
         <input
           {...inputProps}
-          className={joinClassNames(className, formStyles.height, styles.hideNumberSpinner)}
+          className={joinClassNames(
+            className,
+            formStyles.height,
```
theetrain: This can be removed since it sets `height: 3.25rem` that you intend to set to `height: auto` using `styles.input`

PR #717, opened 2018-08-16
docs(core-heading): add rationale for splitting level from style

theetrain:
## Description

* Add more verbose examples for heading levels

## Package changes

```
No changes
```

---

image
image


agorovyi:
Thanks for addressing this! Like the explanation, it makes it more clear why you're allowing this flexibility in styling headings. I guess my only question is to why would a designer choose to visually display H2 followed by H4?

lucylist:
hey @agorovyi a common scenario we're seeing is on marketing pages, where we have a promo banner (first thing top of the page) using the display h1. most of these have text that is promoting the deal. this would be read by the screen reader as an H1, but it's not actually an H1 for the page. it's just a promo. the actual H1 would be PikTV. 

hope that helps...?

screen shot 2018-08-17 at 11 18 00 am

agorovyi:
@lucylist Thanks for the explanation. That helps a lot.

Just one last question. In the example from a screenshot there is an H4 tag (eyebrow styled as H4) that is preceding H1. Wouldn't that break HTML semantics?

lucylist:
hey @agorovyi sorry i missed your question.. those H4 in this case should just be body text bold.. not Headings..

theetrain:
The coded example has been updated. Note the following:

* Our core-display-heading component currently only renders at the 'Display' font size with an `h1` tag. There is no further customization
* The example I've added below uses:
  * core-heading for the eyebrow (first line of text), rendering at the 'h4' font size with an `h1` tag
  * core-heading for the 'Stream live TV...' text, rendering at the 'h1' font size with an `h2` tag
* Only core-heading has the ability to display in the h1-h4 font sizes as either h1-h4 HTML tags. **It is currently not possible for them to be read by assistive technology as anything other than h1 to h4**

image

lucylist:
hey @theetrain as per our discussion in github grooming, we aren't able to specifically say for this example what the titles should be in terms of which heading level. It needs to be in context of the page and what else is above and below it. For this reason, we need to allow teams to be able to adjust the tags to what they need to. I've also confirmed this with @owestinTelus

PR #720, opened 2018-08-22
Update Link.md

jonathanpalma:
I added in a comment the way you should import the TdsLink component considering that when you are using '@tds/core-link' instead of '@telusdigital/tds' there is an error that is not very understandable for most of the new team members that are not familiar with the differences between [default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Importing_defaults) and [named](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Import_a_single_export_from_a_module) imports.

> Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

### DIFF
--- a/packages/Link/Link.md
+++ b/packages/Link/Link.md
@@ -34,6 +34,9 @@ import { Link as ReactRouterLink } from 'react-router'
 
 import { Link as TdsLink } from '@telusdigital/tds'
 
+// If you're importing from @tds instead of telusdigital
+// import TdsLink from '@tds/core-link';
+
 const LinkWrapper = ({ children, ...rest }) => (
   <TdsLink
     {...rest}


theetrain:
Hi @jonathanpalma, thanks for contributing to tds-core.

Seeing that your contribution applies to node module imports in general, and the markdown changes are only going in Link.md, it may not be visible throughout the rest of the documentation site at https://tds.telus.com/components/index.html

Since the scope of this change applies to module imports in general, we recommend following the Reference Architecture guide on transpiling - to use discretion when importing modules and applying aliases to names exports: https://github.com/telus/reference-architecture/blob/master/development/transpiling.md

packages/Link/Link.md
```diff
@@ -34,6 +34,9 @@ import { Link as ReactRouterLink } from 'react-router'
 
 import { Link as TdsLink } from '@telusdigital/tds'
```
theetrain: Please update this example to use `@tds/core-link`

packages/Link/Link.md
```diff
@@ -34,6 +34,9 @@ import { Link as ReactRouterLink } from 'react-router'
...
 
+// If you're importing from @tds instead of telusdigital
```
theetrain: Please remove this commented example.

Issue #509, opened 2018-03-06
[Link] Underline disappears when zoomed out


ryanoglesby08:
## Description

* The "zoom" feature on browsers seems to cause the `Link` component's underline to disappear. It should appear no matter the zoom level.

## Reproduction Steps

(Using Chrome on Mac)

1. View https://tds.telus.com/components/index.html#link
2. Zoom out at least once. Click View -> Zoom Out. Or "Cmd + -"
3. Observe that the underline on all links disappears.

## Reproduction Percentage

* 10/10 times

## Meta

* TDS version: v0.34, 1.0
* High impact: No
* Using Chrome on Mac

PR #722, opened 2018-08-29
fix(core-link): use text-decoration: underline

jraff:
## Related issues

See #509 

## Package changes

```
Changes:
 - @tds/core-link: 1.0.2 => 1.0.3
```

This should be rebased and merged.


packages/Link/Link.modules.scss
```diff
@@ -1,12 +1,11 @@
...
 
 @mixin add-skinny-underline {
-  border-bottom: 0.05em solid currentColor;
-  text-decoration: none;
+  text-decoration: underline;
 }
 
 @mixin clear-skinny-underline {
-  border-bottom: 0.05em solid transparent;
+  text-decoration: none;
```
theetrain: These mixins are only used one time each throughout the codebase. Let's remove these and apply CSS directly to the `.base` class using `:hover`.

PR #739, opened 2018-09-20
feat(core-notification): add dismissible

jraff:
## Description

Allow `<Notification>` to be dismissed.

## Package changes

```
Changes:
 - @tds/core-notification: 1.1.9 => 1.2.0
```


packages/Notification/Notification.jsx
```diff
@@ -31,32 +34,111 @@ const renderIcon = icon => <DecorativeIcon symbol={icon.symbol} variant={icon.co
...
+  }
+
+  adjustContentHeight = () => {
+    if (this.contentWrapper.offsetHeight !== this.state.contentWrapperHeight) {
+      this.setState({ contentWrapperHeight: this.contentWrapper.offsetHeight })
+    }
+  }
+
+  renderNotification() {
+    const { variant, dismissible, dismissibleA11yLabel, children, ...rest } = this.props
```
theetrain: We can add a warning when `dismissible` is defined and `dissmissibleA11yLabel` is undefined, similarly to this:

https://github.com/telus/tds-core/blob/be19a7e7de0790b74c109720edabbe9b29928c3c/packages/ChevronLink/ChevronLink.jsx#L35-L38

packages/Notification/Notification.jsx
```diff
@@ -31,32 +34,111 @@ const renderIcon = icon => <DecorativeIcon symbol={icon.symbol} variant={icon.co
...
+                    isImportant(variant) ? styles.withIcon : styles.withoutIcon
+                  )}
+                >
+                  <Paragraph bold={isImportant(variant)}>{children}</Paragraph>
+                </div>
+              </Box>
+            </FlexGrid.Col>
+          </FlexGrid.Row>
+        </FlexGrid>
+        {dismissible && (
```
theetrain: I'm assuming here: would the dismissible icon be displaced to the far right on desktop viewports? Maybe it's more appropriate to add another `<FlexGrid.Col xs={dismissible ? 1 : 0}>` that wraps the `<StandaloneIcon>`.

jraff: I didn't include the icon in the FlexGrid because the icon is required to be 8px from top and 16px from the right. I think it's better to just absolutely position it there.

PR #808, opened 2018-11-16
docs(core-flex-grid): add docs for spacing between rows

marcod1419:
This adds a section to the `FlexGrid.Row` docs about using `Box` for spacing between rows.

### DIFF
--- a/packages/FlexGrid/Row/Row.md
+++ b/packages/FlexGrid/Row/Row.md
@@ -1,33 +1,45 @@
 ### Alignment
 
-Use flexbox alignment properties to vertically and horizontally align columns.
+Use `FlexGrid` alignment properties to vertically and horizontally align columns.
 
 ```jsx { "props": { "className": "docs_full-width-playground docs_flex-grid-coloring" } }
 <FlexGrid>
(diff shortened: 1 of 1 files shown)


packages/FlexGrid/Row/Row.md
```diff
@@ -78,36 +90,85 @@ Use flexbox alignment properties to vertically and horizontally align columns.
...
 
+### Spacing between rows
+
+Use the [Box](#box) component's `between` prop to add spacing between rows.
+
+_Note:_ When adding spacing between rows, please only use `between` spacing values _5_, _6_ or _7_.
+
+```jsx { "props": { "className": "docs_full-width-playground docs_flex-grid-coloring" } }
+<FlexGrid>
+  <Box between={5}>
```
theetrain: This is improper use. Let's revert it to have `<Box>` nested inside each individual column until we align on a proper solution.

marcod1419:
Because of the design-inconsistent and confusing manner of this solution, this will not be published to our docs and this pull request will be closed. If this issue ever comes up as a support request, please view the Support Strategy doc on our shared drive. A new section called "Approved Workarounds" has been added that has the solution to this included.

PR #824, opened 2018-11-23
Add ButtonGroup Component

marcod1419:
This adds the `ButtonGroup` core component to TDS. It also marks a change in how we'll handle inputs going forward.


packages/ButtonGroup/ButtonGroupItem/ButtonGroupItem.modules.scss
```diff
@@ -0,0 +1,40 @@
+@import '~@tds/core-colours/colours';
```
jraff: Should we be using rem?

packages/ButtonGroup/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@tds/core-button-group",
+  "version": "1.0.0-preview",
+  "description": "",
```
jraff: Need a description here.

packages/ButtonGroup/ButtonGroupItem/ButtonGroupItem.modules.scss
marcod1419: Not sure which line you're referring to here.

packages/ButtonGroup/ButtonGroupItem/ButtonGroupItem.jsx
```diff
@@ -0,0 +1,89 @@
...
+  children,
+  ...rest
+}) => (
+  <div className={styles.itemContainer} {...safeRest(rest)}>
+    <input
+      id={value}
+      name={name}
+      value={value}
+      type="radio"
+      checked={selectedValue === value}
```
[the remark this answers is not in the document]
marcod1419: Not entirely sure what you mean here. Currently, the ButtonGroup does control the selected value through its `value` prop.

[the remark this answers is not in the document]
jraff: I'm thinking about ButtonGroup.Item.

Instead of passing it `selectedValue` and `value` props, can we just pass it a `checked` prop?

Something like this:
```
const getButtons = () => {
    return React.Children.map(children, child =>
      React.cloneElement(child, { name, onChange, onFocus, onBlur, checked: value === child.props.value })
    )
  }
```

packages/ButtonGroup/ButtonGroupItem/ButtonGroupItem.modules.scss
jraff: Everywhere we use px.

marcod1419: Ahh, well here it's ok because they're used for box shadows. The box shadows here were defined as `px`, meaning that they're absolute values. Regardless of zoom level, we want them to keep their pixel count.

packages/ButtonGroup/ButtonGroupItem/ButtonGroupItem.jsx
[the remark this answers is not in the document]
marcod1419: Oh, I see. Sure, I'm good with that.

packages/ButtonGroup/__tests__/ButtonGroup.spec.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+    expect(buttonGroup).toMatchSnapshot()
+  })
+
+  it('selects the correct button', () => {
+    const buttonGroup = doMount(defaultProps)
+    expect(
+      buttonGroup
+        .children()
+        .find('input')
```
theetrain: This use of `.find` finds the first matching element, which happens to be selected button in this test. This test would fail if `value` were to equal `128gb`, even though it may actually be properly selected.

Try:

```jsx
const buttonGroup = doMount({ value="128gb" }) // assumes `defaultProps` are passed

expect(
  buttonGroup
    .children()
    .find('input[value="128gb"]')
    .get(0).props.checked
).toEqual(true)

```

packages/ButtonGroup/__tests__/ButtonGroup.spec.jsx
```diff
@@ -0,0 +1,89 @@
...
+  it('passes additional attributes to the element', () => {
+    const buttonGroup = doShallow({ ...defaultProps, id: 'the-id', 'data-some-attr': 'some value' })
+
+    expect(buttonGroup).toHaveProp('id', 'the-id')
+    expect(buttonGroup).toHaveProp('data-some-attr', 'some value')
+  })
+
+  describe('Accepts A11yContent', () => {
+    it('connects to ButtonGroup.Item', () => {
+      const buttonGroup = mount(
```
theetrain: I think this may work with `shallow`

packages/ButtonGroup/ButtonGroup.md
```diff
@@ -0,0 +1,54 @@
...
+
+The `label` prop on the parent `ButtonGroup` component will add a label to the whole group. This value will be read out loud when the group is selected via a screen reader.
+
+### Usage criteria
+
+- Teams can set a selection to be selected on default
+- Use button group when one selection needs to be made from multiple options
+- Use when there can only be exactly one choice from multiple options
+- Use there are 2-6 options to choose from
+- Use when options are similar in format or category and short (eg. 1GB, 3GB, 5GB)
```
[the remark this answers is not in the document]
marcod1419: I went above and beyond and made it "format or category, and also short".

config/eslint.config.js
```diff
@@ -5,5 +5,13 @@ module.exports = {
     'react/require-default-props': 'warn',
     'import/no-extraneous-dependencies': false, // disabling it because all dev dependencies are pulled into the root
     'prefer-destructuring': 0,
+    'jsx-a11y/label-has-for': [
```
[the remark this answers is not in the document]
marcod1419: Looks like we are using the wrong telus eslint config. We need `@telus/eslint-config` instead of `@telusdigital/eslint-config`. (They're separate packages) The current config is a year out of date. Maybe this should be broken off as an upgrade since this changes our overall tooling? (Also explains why there was no deprecation warning, and why the new rule wasn't found.)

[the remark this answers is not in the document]
theetrain: I've moved this to our backlog 👍

packages/ButtonGroup/ButtonGroup.jsx
```diff
@@ -0,0 +1,100 @@
...
+  })
+
+  if (buttonValues.indexOf(value) === -1) {
+    warn(
+      'ButtonGroup',
+      `Selected value "${value}" of ButtonGroup named "${name}" does not match the value of any button in the group. A button must be selected by default. Available button values are: ${buttonValues}`
+    )
+  }
+  return (
+    <fieldset name={name} {...safeRest(rest)}>
```
theetrain: Though this is inconsequential, we should put `safeRest` before other props for consistency among other components. In the case here, a user cannot override the `name` prop as used here since it matches the `name` html attribute, so `rest` props wouldn't be able to override it.

packages/ButtonGroup/ButtonGroup.md
```diff
@@ -0,0 +1,54 @@
+### Minimal usage
+
+Button Groups are sets of radio inputs styled to look like buttons.
+
+`ButtonGroup` is a stateless component; a button within `ButtonGroup` appears selected when the parent `ButtonGroup` component's `value` prop matches a `ButtonGroup.Item`'s `value` prop. This can be managed by a callback function passed into the `onChange` prop. Additionally, `onFocus` and `onBlur` callback functions are available. All callbacks are assigned to the `ButtonGroup` parent and then passed to all children.
+
+The `label` prop on the parent `ButtonGroup` component will add a label to the whole group. This value will be read by assistive technology when the group is focused.
+
+### Usage criteria
```
theetrain: Move this section to the top of the document, right after the first sentence.

packages/ButtonGroup/__tests__/ButtonGroup.spec.jsx
```diff
@@ -0,0 +1,88 @@
...
+    mount(
+      <ButtonGroup {...defaultProps} {...props}>
+        <ButtonGroup.Item value="64gb">64 GB</ButtonGroup.Item>
+        <ButtonGroup.Item value="128gb">128 GB</ButtonGroup.Item>
+        <ButtonGroup.Item value="256gb">256 GB</ButtonGroup.Item>
+      </ButtonGroup>
+    )
+
+  it('renders', () => {
+    const buttonGroup = doShallow(defaultProps)
```
theetrain: You don't need to pass in `defaultProps` to `doShallow` since it's already there.

packages/ButtonGroup/ButtonGroupItem/ButtonGroupItem.jsx
```diff
@@ -0,0 +1,96 @@
...
+  checked,
+  onChange,
+  onFocus,
+  onBlur,
+  children,
+  ...rest
+}) => {
+  const itemId = generateId(name).postfix(value)
+  return (
+    <div className={styles.itemContainer} {...safeRest(rest)}>
```
theetrain: Reposition `safeRest` to the left

Issue #825, opened 2018-11-23
[Bug] TDS PriceLockup - Component center alignment issues


tarekamin:
## Description

TDS PriceLockup does not centre correctly when wrapped by a centrally aligned component. PriceLockup was implemented in Site Builder - wrapped in a block (BlockPriceLockup). When BlockPriceLockup gets used in a centrally aligned fashion, the top and bottom text get centred while the price, sign, and rate text remain left aligned. This was replicated in the TDS catalogue. 

## Reproduction Steps

- go here https://tds.telus.com/components/index.html#pricelockup
- In the view code section enter this snippet:
```
<FlexGrid>
  <FlexGrid.Row horizontalAlign="center">
    <FlexGrid.Col xs={12} md={12}>
      <PriceLockup
        size="medium"
        topText="Starting at"
        price="25"
        rateText="/month"
        bottomText="$68 /month after 3 months"
        signDirection="left"
      />
    </FlexGrid.Col>
  </FlexGrid.Row>
</FlexGrid>
```

## Meta

- TDS version: v1.0.0
- Willing to develop solution: Willing to get help
- Has workaround: Yes (wrap in left aligned components)
- High impact: No

## Screenshots

![image](https://user-images.githubusercontent.com/29206760/48962130-cfe28e00-ef30-11e8-8816-49c8f06f11ff.png)

nicmak:
@tarekamin  Thanks, I will be looking into this!

tarekamin:
Thanks @nicmak! Let me know if you need further support.

theetrain:
This issue can also be tracked on Jira: https://telusdigital.atlassian.net/browse/TDS-1047

jraff:
@tarekamin fixed with [@tds/core-pricelockup@1.0.3](https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-price-lockup%401.0.3)

tarekamin:
Thanks so much folks! Verified it behaves correctly on Site Builder.

PR #836, opened 2018-11-26
fix(core-price-lockup): fixing horizontal alignment issue

nicmak:
## Related issues

https://github.com/telus/tds-core/issues/825

## Description

- wrapped a div with inline-block and alignment left to account FlexGrid Row horizontal alignment changes

- [x] New code is unit tested
- [x] For code changes, run `yarn prepr` locally


packages/PriceLockup/PriceLockup.modules.scss
```diff
@@ -1,6 +1,11 @@
 @import '~@tds/core-responsive/responsive';
 @import '../../shared/scss/typography';
 
+.wrapperAlignment {
+  text-align: left;
+  display: inline-block;
```
theetrain: This fix can be achieved without `display: inline-block`

image

packages/PriceLockup/PriceLockup.jsx
```diff
@@ -67,31 +67,33 @@ const PriceLockup = ({ size, price, topText, signDirection, rateText, bottomText
...
-          {renderPriceValueSign()}
-          {rateText && (
-            <span
-              data-testid="rateText"
-              className={size === 'large' ? styles.largeText : styles.mediumText}
-            >
-              {rateText}
-            </span>
-          )}
+    <div className={styles.wrapperAlignment}>
```
theetrain: Rather than add a wrapping `<div>`, let's make use of the Box's `dangerouslyAddClassName` on line 71.

packages/PriceLockup/PriceLockup.modules.scss
nicmak: If i remove inline-block (this is when I include the code inside box component), horizontal alignment does not work from FlexGrid.Row.

theetrain: If we release this using `inline-block` then it will be a breaking change because it would cause subsequent inline elements to appear inline, for example:

image

```jsx
<PriceLockup
  size="medium"
  topText="Starting at"
  price="25"
  rateText="/month"
  bottomText="$68 /month after 3 months"
  signDirection="left"
/>
<ButtonLink href="#">Click me</ButtonLink>
```

We have backlogged a patch to have Price Lockup have a max width, that way the hairline divider's width won't need to be controlled by the developer via FlexGrid: https://telusdigital.atlassian.net/browse/TDS-1042

theetrain:
Thanks for the update, please address the following:
- Remove `inline-block` (see comment above)
- Update Jest snapshot and e2e screenshot for the build to pass

Issue #746, opened 2018-10-04
Radio component with label and description


ivaylogabrovsky:
## Description

* Description of the problem
We need Radio component with short label and description aligned to label in different resolutions.
## Recommendation

* Description of the proposed solution
<Radio
                  label={this.getTitle(this.pageLabels.options.manager.title)}
                  name="manager"
                  value={PROFILE_TYPE.MANAGER}
                  checked={this.state.choice === PROFILE_TYPE.MANAGER}
                  onChange={this.setChoice}
                />
<Responsive minWidth="md">
                  {matches => (
                    <Box horizontal={matches ? 4 : 5}>
                      <Text>{this.pageLabels.options.manager.subTitle}</Text>
                    </Box>)
                  }
</Responsive>

## Designs

* Screenshots

![screen shot 2018-10-04 at 3 02 43 pm](https://user-images.githubusercontent.com/30123542/46497193-dfbedb00-c7e7-11e8-8d7e-cc44d266f4c7.png)
![screen shot 2018-10-04 at 3 02 23 pm](https://user-images.githubusercontent.com/30123542/46497199-e3eaf880-c7e7-11e8-9267-e88ee1b09255.png)

lucylist:
spoke to cherry today. she will add the new designs/screenshots to the ticket.

cherryhung:
<img width="300" alt="screen shot 2018-11-07 at 2 45 39 pm" src="https://user-images.githubusercontent.com/40604656/48156711-ed122e00-e29b-11e8-88a3-19e1db13a485.png">

Updated with new designs. Screenshot and sketch file attached. 

[Form_Radio+Description.zip](https://github.com/telus/tds-core/files/2559000/Form_Radio%2BDescription.zip)

lucylist:
Awesome. thanks @cherryhung !! thinking that when we contribute the designs we might have to move the radio text up a bit to centre align with the circle.
@gkychow good?

cherryhung:
Updated! 
<img width="581" alt="screen shot 2018-11-21 at 1 26 46 pm" src="https://user-images.githubusercontent.com/40604656/48862060-2b771500-ed93-11e8-94d6-b43771f5f3f0.png">
[Form_Radio+Description.zip](https://github.com/telus/tds-core/files/2605406/Form_Radio%2BDescription.zip)

marcod1419:
This has been added to the latest version of Radio! @ivaylogabrovsky @cherryhung

PR #847, opened 2018-12-04
Add description text functionality to core-radio & shared-choice

marcod1419:
See #746

This PR adds description functionality to core-radio. It allows a string to be passed into a new `description` prop that will generate a line of descriptive text below the radio input.


packages/Radio/Radio.md
```diff
@@ -43,3 +44,41 @@ const setChoice = event => {
   />
 </Box>
 ```
+
+### Radio descriptions
+
+In special circumstances, it may be impossible for the radio label to be descriptive enough for the user to make an informed choice. In these cases, use the `description` prop. This prop adds a string of text under a particular radio input that further describes its choice.
```
theetrain: > This prop `can be used to add descriptive text` under a particular radio input...

shared/components/Choice/Choice.modules.scss
```diff
@@ -78,3 +78,8 @@
   position: absolute;
   top: auto;
 }
+
+.descriptionText {
+  margin-top: 0.5rem;
```
theetrain: Should this be managed by Box?

shared/components/Choice/Choice.jsx
```diff
@@ -49,8 +49,13 @@ const renderColoredLabel = (color, content) => (
   </ColoredTextProvider>
 )
 
-const renderLabel = (label, feedback, checked, disabled) => {
-  const content = <Text size="medium">{label}</Text>
+const renderLabel = (label, description, feedback, checked, disabled) => {
```
theetrain: Adding the `description` prop would technically make this available for core-checkbox. Should we add a separate story to work on adding description text for core-checkbox?

marcod1419: I think there's potential for that, but we should consult design.

jraff: Good catch, I think it's OOS for this fix. We can always add this as a feature to @tds/core-checkbox in the future.

Issue #731, opened 2018-09-11
Steptracker UI Discrepancy


cherryhung:
## Description

* There is a discrepancy between the steptracker UI on DSM and in code. On DSM, there is a grey background while tds.telus.com has a white background and divider line. 

## Reproduction Steps

* N/A

## Reproduction Percentage (if applicable)

* N/A

## Meta

* TDS version: v1.0.5
* Willing to develop solution: Yes
* Has workaround: No
* High impact: Yes

## Screenshots

* Include any relevant screenshots
<img width="717" alt="dsm" src="https://user-images.githubusercontent.com/40604656/45371287-386bd100-b5b8-11e8-8790-c099d5e12035.png">
<img width="1093" alt="tdscom" src="https://user-images.githubusercontent.com/40604656/45371294-3dc91b80-b5b8-11e8-9a85-c892f80101e0.png">

lucylist:
hey @cherryhung thanks for noting this! It looks like the component is able to have a grey background and not.
we will match the tds docs example to match with the DSM component and also provide guidelines on usage

theetrain:
Next steps: determine whether StepTracker should be a full-width component (like Notification), or should optionally pass in a prop to make its background Athens Grey.

theetrain:
We will enhance the StepTracker to be a full-width component with a default background of Athens Grey that can be optionally set to white (not transparent).

theetrain:
Update: this story is being tracked in https://telusdigital.atlassian.net/browse/TDS-985

marcod1419:
This is added in the latest version of StepTracker! @cherryhung

PR #858, opened 2018-12-10
feat(core-step-tracker): make full width, add bg colour options

marcod1419:
See: #731

This makes step tracker full width while also making the background colour grey by default. It includes a new backgroundColour prop to switch the background to white if preferred.


packages/StepTracker/package.json
```diff
@@ -27,7 +27,8 @@
   "dependencies": {
     "@tds/core-box": "^1.0.1",
     "@tds/core-decorative-icon": "^1.0.0",
-    "@tds/core-text": "^1.0.3",
+    "@tds/core-flex-grid": "^2.3.0",
+    "@tds/core-text": "^1.0.2",
```
jraff: It looks like core-text was downgraded.

packages/StepTracker/StepTracker.modules.scss
```diff
@@ -1,10 +1,29 @@
...
+  flex-direction: row;
+}
+
+.greyBg {
+  background-color: $color-athens-grey;
+  @include step-tracker-padding;
+}
+
+.whiteBg {
+  background-color: #ffffff;
```
jraff: Should we use `$color-white` here?

marcod1419:
@jraff It is. Just forgot to add the indicator! I'll do that now.

packages/StepTracker/package.json
marcod1419: Weird. Will fix.

packages/StepTracker/StepTracker.modules.scss
marcod1419: Sure.

Issue #879, opened 2018-12-31
[ExpandCollapse] - calling setState in child component causes re-render


theetrain:
## Description

Originally reported by @nicmak 

When a child component within `<ExpandCollapse.Panel />` calls `this.setState()`, the parent `<ExpandCollapse />` component re-renders and returns to its default state, which is collapsed. 

## Reproduction Steps

1. Using the following code sample:

```jsx
class MyComponent extends React.Component {
  constructor(props) {
    super(props)

    this.changeHandler = this.changeHandler.bind(this)
  }

  state = {
    checked: false
  }

  changeHandler() {
    this.setState((previous) => {
      return { checked: !previous.checked }
    })
  }

  render() {
    return (
      <ExpandCollapse tag="h2">
        <ExpandCollapse.Panel id="features" header="Features">
          <Checkbox
            label="Hello"
            name="hello"
            value="Hii"
            checked={this.state.checked}
            onChange={this.changeHandler}
          />
        </ExpandCollapse.Panel>
      </ExpandCollapse>
    )
  }
}
```

2. Render in a React application
3. Click on the `<Checkbox />` component
4. Observe the `<ExpandCollapse.Panel />` collapses

## Reproduction Percentage (if applicable)

- 10/10 times

## Meta

- TDS version: v1.2.3
- Willing to develop solution: Yes, by @nicmak 
- Has workaround: No
- High impact: No

## Screenshots

![screen recording 2018-12-31 at 02 43 pm](https://user-images.githubusercontent.com/12798751/50566686-7c6f1700-0d0a-11e9-8720-c7be33eb175f.gif)

ah-arch:
ExpandCollapse’s `componentWillReceiveProps` is not triggered by PanelWrapper’s `onClick/togglePanel`, but `this.setState` does. That resets the `openPanels` back to what `props.open` was set to, which was originally nothing:
https://github.com/telus/tds-core/blob/727f5211fabc00554573bf61e5a1863d25462244/packages/ExpandCollapse/ExpandCollapse.jsx#L22-L23

Perhaps if `nextProps.open` differs with  `props.open`, then the `state.openPanels` should be modified to include or exclude the only the **changed** `nextProps.open`'s `panelId`s, e.g.:
https://github.com/telus/tds-core/blob/727f5211fabc00554573bf61e5a1863d25462244/packages/ExpandCollapse/ExpandCollapse.jsx#L37-L41

simpleimpulse:
[...]

PR #887, opened 2019-01-04
fix(core-expand-collapse): fixing expand collapse defect with Checkbox

nicmak:
#879

## Description
- the code makes your that the ExpandCollapse component does not re-render

### DIFF
--- a/packages/ExpandCollapse/__tests__/ExpandCollapse.spec.jsx
+++ b/packages/ExpandCollapse/__tests__/ExpandCollapse.spec.jsx
@@ -216,11 +216,14 @@ describe('ExpandCollapse', () => {
     it('triggers individual panel callbacks when panels are opened and closed', () => {
       const onPanelToggle = jest.fn()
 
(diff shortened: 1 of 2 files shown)


packages/ExpandCollapse/__tests__/ExpandCollapse.spec.jsx
```diff
@@ -216,15 +216,12 @@ describe('ExpandCollapse', () => {
...
       const onPanelToggle = jest.fn()
 
-      const { expandCollapse, clickPanel } = doMount(
+      const { clickPanel } = doMount(
         <ExpandCollapse>{aPanel({ id: 'panel-1', onToggle: onPanelToggle })}</ExpandCollapse>
       )
 
       clickPanel('panel-1')
       expect(onPanelToggle).toHaveBeenCalledWith(true)
-
```
theetrain: Why was this test removed? With this change, this only tests whether callbacks are called when panels are opened.

PR #902, opened 2019-01-25
Build pipeline updates for OpenShift 3.11

simpleimpulse:
See: [telus/technology-forum#266](https://github.com/telus/technology-forum/issues/266)


marcod1419:
Once [this](https://github.com/telus/tds-core/pull/918) gets pulled into master, this PR should be rebased so it can be properly tested with the new changes.

openshift/run-deploy-docs.sh
```diff
@@ -40,3 +40,5 @@ oc run ${NAME}-${VERSION} \
       }]
     }
   }'
+
+oc rollout status dc/telus-isomorphic-starter-kit-${ENVIRONMENT} --watch
```
jraff: I don't think this should be `dc/telus-isomorphic-starter-kit-${ENVIRONMENT}`

PR #911, opened 2019-01-31
chore(other): add deploy steps to circleci

marcod1419:
This PR adds deploy steps to our CircleCI script. Due to the nature of this change, we need to QA on master after fully verifying we are confident in the presented changes.


circle.yml
```diff
@@ -94,9 +142,16 @@ workflows:
...
+      - approve-release:
+          type: approval
           requires: [build, lint, unit, e2e]
           filters:
             branches:
               only: master
+
+      - release:
+          context: npm-library
+          requires: [approve-release]
```
theetrain: Perhaps instead of requiring `approve-release`, you can remove the approve-release step and add `requires` and `filters` to this step.

theetrain: Update: it turns out this is necessary.

PR #916, opened 2019-02-04
Docker e2e fixes

marcod1419:
This fixes some issues with the new e2e implementation.

- The test will now once again test only packages that were updated by default.
- The "cannot write to chromelog" error will no longer appear at the end of tests.
- The e2e startup message has been modified to indicate that no action is required from the user for it to start.


jraff:
r+, continues to look good.

Issue #894, opened 2019-01-17
[Input] The error prop does not allow links


theetrain:
## Description

Originally reported by @ah-arch 

Defect: When making use of the core-input component's `error` prop, I pass paragraph text with a core-link component, then a prop-type error appears since `error` only accepts strings.

Expectation: I can pass paragraph text and a core-link component without receiving a prop-type error.

Note, this defect may also be applicable to other form components.

## Reproduction Steps

1. Within an application's development environment, insert an Input component that makes use of the `error` prop like this:
   ```jsx
    <Input label="First name" value="Harry" error={(
      <Paragraph>Hello <Link href="https://google.com">What is this</Link></Paragraph>
    )} />
   ```
2. Observe a propType error with for the `error` prop within the FormField component

## Reproduction Percentage (if applicable)

- 10/10 times

## Meta

- TDS component version: @tds/core-input@1.0.13 
- Willing to develop solution: Yes @ah-arch 
- Has workaround: Yes 
- High impact: No 

## Screenshots

![image](https://user-images.githubusercontent.com/12798751/51347281-a72ccf80-1a6d-11e9-9568-a199c61e3569.png)

ah-arch:
The `Notification`component already automatically adds a `p` tag, and therefore it would be invalid to have a nested `p` tag in the `p` tag, or `div`s.
So the only allowed elements are really `span` or `<React.Fragment>{string} <Link /></React.Fragment>`
This is our scenario using a `Fragment`:
![error with link](https://user-images.githubusercontent.com/4450690/51347592-8ca72600-1a6e-11e9-89c4-6d6f57f7ecfa.png)

Christina-Lo:
There is consensus that this is a useful pattern from the DPA meeting and we will move forward with implementing this

ah-arch:
Realized that the `error` prop was not placed in a `Notification`, but an `InputFeedback` with a nested `Paragraph` instead.

Leaving it as it, and just updating `error` to `PropTypes.oneOfType([PropTypes.string, PropTypes.node])`.

marcod1419:
@ah-arch That sounds good! Just one thing, `PropTypes.node` could be `componentWithName(Link)`. If you're not aware of it, we recently created a [PropTypes utility](https://tds.telus.com/components/index.html#proptypes) so we have a lightweight way of being more specific with our props. Thanks so much!!

ah-arch:
Most likely the node will contain a combination of strings and `Link`.

[...]

Issue #893, opened 2019-01-15
[Input] optionally display hint text below label


theetrain:
## Description

Originally reported by Apurv and @ah-arch.

When using the TDS Input component, my design calls for more hint text than what is able to comfortably fit within a small parent container. 

## Recommendation

Provide a means for designers to optionally provide hint text below the Input's label, rather than inline.

TDS recommendation: add a new `description` prop that adds text below the label, and keep `hint` available for brief descriptive messages that display inline with the label. `description` would have a similar behaviour to core-radio's `description` prop.

Inline text (`hint`) and descriptive text (`description`) must not be used together. Only one or neither can be present in a given time.

## Designs

This has been approved by design leads.

![image](https://user-images.githubusercontent.com/12798751/51206798-fd1b3f00-18d6-11e9-8deb-41feb72679cf.png)

## Meta

- TDS component: @tds/core-input@1.0.13
- Willing to develop solution: Yes
- Has workaround: No 
- High impact: Yes 
- Designer/Developer Pair: Apurv Ray, @ah-arch

ah-arch:
Can a user supply both a `description` and a `helper` or `hint`?
Where would the `helper` go - below or above the `description`?

theetrain:
> Can a user supply both a description and a helper or hint?

We'll tackle this design problem in (issue number)

> Where would the helper go - below or above the description?

Here's an alternative prop interface suggested by @jraff: `hintPosition: PropTypes.oneOf(['inline', 'below'])` in which you can manually set the hint position to be inline or below, relative to the input label.

As a future enhancement, we could even add a 'responsive' prop value, which could work like this:

```jsx
<Input 
  label="First name"
  value="Enrico"
  hint="Your first name"
  hintPosition="responsive" />
// change hint text position automatically based on parent width and character count

<Input
  label="First name"
  value="Enrico"
  hint="Your first name"
  hintPosition={{xs: 'below', md: 'inline'}} />
// change hint text position based on breakpoint using responsiveProps
// See @tds/util-prop-types https://tds.telus.com/components/index.html#proptypes
```

> And this prop should also apply to TextArea and Select?

Correct, since core-input is a higher order component of FormField, as are other TDS form components, these patterns should apply to all other form components as well.

ah-arch:
Is this wording okay for **Usage Criteria**?
> The `hint` should not be more than 85 characters when positioned below the label

theetrain:
[...]

PR #920, opened 2019-02-05
feat(core-input): TDS-1128

ah-arch:
## Related issues

See #893 and #894, and [TDS-1128](https://telusdigital.atlassian.net/browse/TDS-1128).

## Description

- Adds `hintPosition` to align `hint` below the `label`
- Adds `node` to `error` propTypes to allow strings with `Link`s

- [x] New code is unit tested
- [x] Commits follow our [Developer Guide](https://tds.telus.com/contributing/developer-guide.html#make-a-commit)
- [x] For code changes, run `yarn prepr` locally


marcod1419:
Hi @ah-arch! Just wanted to let you know that we see your pull request and we'll review it as soon as we are able. Thanks for contributing!

packages/Input/Input.md
```diff
@@ -89,13 +89,17 @@ that the field is optional. It is a more usable and accessible option than the H
...
+<Input
+  label="Quick view"
+  hintPosition="below"
+  hint="Enter the account number in the box below to go to the overview page for that account."
+/>
 ```
+
+### Usage criteria
+
+- The `hint` should not be more than 85 characters when positioned below the label
```
simpleimpulse: Consider adding more comprehensive usage criteria, recommending for what character lengths should be using 'inline' vs 'below'.

packages/Input/__tests__/Input.spec.jsx
```diff
@@ -69,6 +69,20 @@ describe('Input', () => {
...
+    it('can have a long hint', () => {
+      const { input } = doMount({
+        id: 'the-id',
+        hint: 'The long hint over here',
+        hintPosition: 'below',
+      })
+
+      expect(input).toContainReact(
+        <Paragraph id="the-id_hint" size="medium">
+          The long hint over here
```
simpleimpulse: Consider using an example of a hint that is so long that it can't be used as an inline hint.

ah-arch: Will find 85 characters of lorem ipsum!

packages/Input/Input.md
ah-arch: Is there a character cut-off for using `inline` vs `below`? What should be recommended?

packages/Input/__tests__/Input.spec.jsx
simpleimpulse: You're right that this doesn't make any difference while programmatic testing, so we can safely skip this part.

packages/Input/Input.md
ah-arch: Updated based on conversation on TDS-1128

packages/Input/Input.jsx
```diff
@@ -60,7 +64,7 @@ Input.propTypes = {
   /**
    * An error message. Either an error or a helper should be used, not both.
    */
-  error: PropTypes.string,
+  error: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
theetrain: Let's change `PropTypes.node` to `componentWithName('Link')` to restrict this to text and the core-link component.

packages/Input/Input.md
```diff
@@ -89,13 +89,20 @@ that the field is optional. It is a more usable and accessible option than the H
...
-)
-;<Input label="Credit Card Number" helper={creditCards} />
+<Input
+  label="Quick view"
+  hintPosition="below"
+  hint="Enter the account number in the box below to go to the overview page for that account."
+/>
 ```
+
+### Usage criteria
```
theetrain: Please move the usage criteria section to the top of this document.

packages/Input/Input.md
```diff
@@ -89,13 +89,20 @@ that the field is optional. It is a more usable and accessible option than the H
...
+
+- Use `hintPosition` to re-position the `hint` below the `label` if the inline `hint` wraps
+  - The `hint` should **not** be longer than 85 characters
+  - The `hint` should **not** wrap more than 3 lines on mobile
```
theetrain: I'm not 100% sure here, but perhaps it should not wrap more than 3 lines `on all viewport sizes`

shared/components/FormField/FormField.jsx
```diff
@@ -170,9 +185,10 @@ class FormField extends React.Component {
 FormField.propTypes = {
   label: PropTypes.string.isRequired,
   hint: PropTypes.string,
+  hintPosition: PropTypes.oneOf(['inline', 'below']),
   value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
   feedback: PropTypes.oneOf(['success', 'error']),
-  error: PropTypes.string,
+  error: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
theetrain: Use `componentWithName` as mentioned above.

shared/components/FormField/__tests__/FormField.spec.jsx.md
```diff
@@ -1 +1 @@
-// FormField is used by Textarea and all functionality is tested through it.
+// FormField is used by Textarea and Input, and all functionality is tested through them.
```
theetrain: Good catch! I poked around and noticed `Select` also uses FormField.

ah-arch: Will update to:
> FormField is used by Input, Textarea and Select, and all functionality is tested through them.

packages/Input/Input.md
ah-arch: Will update to:
>  The `hint` should **not** wrap more than 3 lines on all viewport sizes

packages/Input/Input.jsx
ah-arch: But we need to allow a combination of strings **and** Link: https://github.com/telus/tds-core/issues/894#issuecomment-455325223

Seems like [componentWithName](https://github.com/telus/tds-core/blob/367747116d1afcd7b3be304db84c7846093c267e/packages/prop-types/PropTypes.md#componentwithnamestring) would force the whole content to be a `Link`

This could be updated to use `or` though:

> `or([PropTypes.string, PropTypes.node])`

packages/Input/Input.md
ah-arch: Will move it below the `Minimal usage` section

shared/components/FormField/FormField.jsx
ah-arch: Leaving it due to reason mentioned [above](https://github.com/telus/tds-core/pull/920#discussion_r259396513).

ah-arch:
> Changes:
>  - @tds/core-input: 2.0.0 => 2.1.0
>  - @tds/core-select: 2.0.0 => 2.0.1
>  - @tds/core-text-area: 2.0.0 => 2.0.1
>  - @tds/shared-form-field: 2.0.0 => 2.1.0 (private)

Seems like we are at `2.0.0` without deprecating the `helper`. Is this description still valid?
https://github.com/telus/tds-core/blob/8fa6173f750e7140cbd5ffc1c0d52a75d26c7e64/packages/Input/Input.jsx#L77-L78

packages/Input/Input.jsx
```diff
@@ -70,6 +74,9 @@ Input.propTypes = {
    *
    * **Deprecation:** As of v1.0.0, passing `Input.Helper` will no longer work.
    *
+   * **Deprecation:** As of v2.0.0, it will no longer exist, as it is not a recommended pattern.
```
theetrain: This appears to be a moving target on our part. We sunset features based on adoption, not version. Let's rephrase this to:
> This is not a recommended pattern and will be removed in a future release.

packages/Input/Input.jsx
theetrain: Can we use `componentWithName('Link')` here and in FormField?

ah-arch: Ok, will update to:
> PropTypes.oneOfType([PropTypes.string, componentWithName('Link')])

ah-arch: Updating to `error: or([PropTypes.string, componentWithName('Link')])` will cause the following warning:
error

When used with the following examples in https://github.com/telus/tds-core/issues/894, e.g.:
```<Input error={<React.Fragment>Hello <Link href="#">What is this</Link></React.Fragment>}  />```

Are we not allowing this format of messages?

File: packages/Input/Input.md [hunk 9]
```diff
@@ -14,12 +14,19 @@ For information on how to use disabled Inputs, please refer to the [disabled for
...
 </Box>
 ```
 
+### Usage criteria
+
+- Use `hintPosition` to re-position the `hint` below the `label` if the inline `hint` wraps
+  - The `hint` should **not** be longer than 85 characters
+  - The `hint` should **not** wrap more than 3 lines on all viewport sizes
+- We recommend `React.Fragment` when supplying custom strings to the `error` prop
+  - Note that the `div` tag is not valid in a `p` tag
```
theetrain: Let's add some more notes here:

- Keep `error` text as brief as possible, should be limited to text and links

packages/Input/Input.jsx
theetrain: Please rewrite this as `error: PropTypes.node,` as we'll accept all children for now, combined with additional usage criteria to guide designers and developers on their use of the `error` prop.

jraff:
Closing, see https://github.com/telus/tds-core/pull/952 for new PR.

Issue #925, opened 2019-02-08
TDS/Input does not accept undefined value


ElliottJRo:
## Description

Attempting to use HTML attr `defaultValue` to treat `@tds/core-input` as an [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html#default-values) but unable.

## Reproduction Steps

```jsx
_onChangeHandler = (val) => {
  this.props.updateDataAction(val)
}
render () {
<Input
  onChange={this._onChangeHandler}
  value={this.props.value}
/>
}
```

Expected input value to not change after user types and `_onChangeHandler` sanitizes the input and saves in redux store. `this.props.value` comes from redux.

After finding out how to use defaultValues, I modified render to the following:
```jsx
_onChangeHandler = (val) => {
  this.props.updateDataAction(val)
}
render () {
delete Input.defaultProps.value
<Input
  label={label}
  defaultValue={this.props.defaultValue}
  onChange={this._onChangeHandler}
  value={undefined}
/>
}
```

Expected input value to be set once and not change on subsequent renders and not throw any errors. Actual results: default value was not set, tds/shared/formField threw error saying `value` is required.

## Meta

- TDS component version: @tds/core-input: 1.0.13
- Willing to develop solution: Yes
- Has workaround: No (potential workaround is to use vanilla html input and copy tds/core-input styles. Very messy)
- High impact: Yes

## Notes:
Looks like we need to do the following:
1) Remove defaultProp empty string for `tds/core-input` prop `value`
2) Modify [`FormField`](https://github.com/telus/tds-core/blob/master/shared/components/FormField/FormField.jsx) to not require prop `value`
3) Make sure any components using `FormField` does not need `value` to be required || empty string

ElliottJRo:
Fixed in release @tds/core-input@2.0.0

PR #936, opened 2019-02-12
feat(core-input): allow uncontrolled component usage via tds

ElliottJRo:
## Related issues

See #925 

## Description

At the moment, core-input fails to allow `defaultValue` prop usage. This is due to `Input.defaultProps.value === ''`. Another issue was `FormField` passing down value prop onChange event.

The following was done to allow [React uncontrolled components](https://reactjs.org/docs/uncontrolled-components.html) usage:

- Changed `shared-form-field` to not handle changes (except call onChange) if `defaultValue` is passed in
- Changed `core-input`, `core-select`, `core-text-area` defaultProps.value to undefined
- Refactored `core-select` so it can be used as uncontrolled component (or not)
- Updated test cases to reflect code changes
- Updated docs

- [x] New code is unit tested


ElliottJRo:
Note: I was unable to run `npm run prepr` for some reason

jraff:
Hi @ElliottJRo, what do you think about implementing FormField like this https://gist.github.com/jraff/1a049d694a6947151024eb20d5223ba4#file-formfield-jsx-L80? Although not a truly uncontrolled component, it would appear that way to the user with state being managed purely within the component. The only change we make is to check if `this.props.defaultValue` exists in `componentDidMount` and initialize state then.

PR #948, opened 2019-02-22
Styled/responsive

jraff:
Add named exports to core-responsive to support media queries generation in js


packages/Responsive/Responsive.jsx
```diff
@@ -4,12 +4,7 @@ import Media from 'react-media'
 
 import { warn } from '../../shared/utils/warn'
 
-const breakpoints = {
-  sm: 576,
-  md: 768,
-  lg: 992,
-  xl: 1200,
-}
+import breakpoints from './breakpoints'
```
theetrain: As a suggestion, you can always export small amounts of code without needing to create new files. 

```diff
- const breakpoints = {
+ export const breakpoints = {
  sm: 576,
  md: 768,
  lg: 992,
  xl: 1200,
}
```

packages/Responsive/breakpoints.js
```diff
@@ -0,0 +1,6 @@
+export default {
```
theetrain: This file can be removed if exporting code from Responsive.jsx

packages/Responsive/media.js
```diff
@@ -0,0 +1,9 @@
+import breakpoints from './breakpoints'
+
+export default {
+  from: breakpoint => style => ({
+    [`@media (min-width: ${breakpoints[breakpoint]}px)`]: {
```
theetrain: Also, should this support `max-width` just like our `sass-mq` library did? Am I missing some other use case here?

jraff: `min-width` should handle almost all cases right now. We can always enhance media to include `max-width` later.

packages/Responsive/Responsive.md
```diff
@@ -83,8 +83,55 @@ To implement responsive behaviour in your component, this package provides Sass-
 
 The `Responsive` React component is a thin wrapper over the [**react-media**](https://github.com/ReactTraining/react-media) community component, which "listens for matches to a CSS media query and renders its child component based on whether the query matches or not." While it is available, this component is not recommended for most use-cases.
 
+### `media` utility usage
+
+To generate media queries for your `styled-components`, first import the named `media` export from `@tds/core-colours` in your .`jsx` file.
+
+```js static
+import { media } from '@tds/core-colours'
```
theetrain: Import from `@tds/core-responsive`

packages/Responsive/Responsive.md
```diff
@@ -1,18 +1,18 @@
...
+The package also contains a `media` utility (documented below). In most cases **we recommend using the media utility over the Responsive React component** since they do not adversely affect server-side rendering. While not recommended for general use, the Responsive component is useful if you want to do something programmatically. For example, changing a prop or swapping a component on resize.
 
 ### Breakpoints
 
-The included Sass-based media queries and the `Responsive` component both use the same breakpoint keys. The following chart details these breakpoint keys, and what values they represent.
+The included `media` utility and the `Responsive` component both use the same breakpoint keys. The following chart details these breakpoint keys, and what values they represent.
 
-| Breakpoint key | `min-width` / `$from` Value | `max-width` / `$until` Value |
```
theetrain: Please copy this table to the Sass documentation below, that way consumers can still reference the `$from` and `$until` values.

packages/Responsive/__tests__/media.spec.js
```diff
@@ -0,0 +1,37 @@
+import media, { breakpoints } from '../media'
+
+describe('media', () => {
+  describe('from()', () => {
```
theetrain: Please add tests for other functionalities: until, and, css.

PR #952, opened 2019-02-26
feat(core-input): TDS-1128

jraff:
See https://github.com/telus/tds-core/pull/920


jraff:
Hey @ah-arch, according to design there should not be any space between the label and the hint text when the hintPosition is "below".

ah-arch:
Will un-Box it. @jraff

shared/components/FormField/FormField.jsx
```diff
@@ -133,25 +149,29 @@ class FormField extends React.Component {
...
-        <Flexbox
-          direction="row"
-          justifyContent="spaceBetween"
-          dangerouslyAddClassName={positionStyles.relative}
-        >
-          {renderLabel(label, hint, fieldId)}
-
-          {tooltip && React.cloneElement(tooltip, { connectedFieldLabel: label })}
-        </Flexbox>
+        {hintId ? renderHintBelowLabel(labelWrapper, hintId, hint) : labelWrapper}
```
ah-arch: @jraff made the change here.

jraff:
r+

packages/Input/Input.md
```diff
@@ -14,12 +14,20 @@ For information on how to use disabled Inputs, please refer to the [disabled for
 ```jsx
 <Box between={2}>
   <Input label="First name" value="Harry" />
-
   <Input type="tel" label="Mobile phone" pattern="[0-9]{10}" maxLength="10" />
   <Input type="password" label="Tax ID" value="123456789" autoComplete="on" />
 </Box>
 ```
 
+### Usage criteria
```
Christina-Lo: Let's have usage criteria under every major feature and please add the following regarding hint text:
Hint can now be used inline (beside) and below the label

ah-arch: What does "under every major feature" mean?

So the points should be updated to this?
> - Hint can now be used inline (beside) and below the label
> - Use `hintPosition` to re-position the `hint` below the `label` if the inline `hint` wraps
>   - The `hint` should **not** be longer than 85 characters
>   - The `hint` should **not** wrap more than 3 lines on all viewport sizes

Christina-Lo: Maybe they're not major features? But I meant these ones under forms:
Button, Input, Checkbox, Radio, ButtonGroup, Select, TextArea, InputFeedback, Tooltip

ah-arch: Should another ticket be used to discuss the scope of moving _Usage Criteria_ for all or some components to the top of each `readme.md` ?

PR #957, opened 2019-02-28
Publishing shared animation

Andrew-K-Lam:
## Description
Publishing shared animation

- [x] New code is unit tested


scripts/build.js
```diff
@@ -27,16 +27,15 @@ getPackageNames(packageNames => {
       'exec',
       '--scope',
       scopeGlob,
-      '--ignore',
-      '@tds/shared-*',
+      '--no-private',
```
theetrain: Will this no longer create version bumps for private packages?

jraff: https://www.npmjs.com/package/@lerna/filter-options#--no-private

`@tds/shared-*` components used to be private packages that were used in tds-core.

If my understanding is correct, now that we are publishing components with the `@tds/shared-*` prefix we can no longer ignore them, instead we can use [`--no-private`](https://www.npmjs.com/package/@lerna/filter-options#--no-private) to exclude all private packages.

packages/Select/__tests__/Select.spec.jsx
```diff
@@ -6,9 +6,9 @@ import Text from '@tds/core-text'
 import Paragraph from '@tds/core-paragraph'
 import InputFeedback from '@tds/core-input-feedback'
 import DecorativeIcon from '@tds/core-decorative-icon'
+import { Fade } from '@tds/shared-animation'
```
theetrain: Was this reflected in its package.json?

packages/TextArea/__tests__/TextArea.spec.jsx
```diff
@@ -5,9 +5,9 @@ import StandaloneIcon from '@tds/core-standalone-icon'
 import Text from '@tds/core-text'
 import Paragraph from '@tds/core-paragraph'
 import InputFeedback from '@tds/core-input-feedback'
+import { Fade } from '@tds/shared-animation'
```
theetrain: Was this reflected in its package.json?

scripts/build.js
theetrain: I see, so this only excludes private packages from **build**, and none of them have a rollup config, so they wouldn't build anyway.

packages/Select/__tests__/Select.spec.jsx
Andrew-K-Lam: Yes: https://github.com/telus/tds-core/blob/2ae6bdb60a46f22854c5ee965379bf0597bdb640/packages/Select/package.json#L38

packages/TextArea/__tests__/TextArea.spec.jsx
Andrew-K-Lam: Yes https://github.com/telus/tds-core/blob/4e1224d72a56a3b0a140b62c27fbcd82f0612add/packages/TextArea/package.json#L37

packages/Select/__tests__/Select.spec.jsx
theetrain: From this commit hash, I see that it is still in `devDependencies`

https://github.com/telus/tds-core/blob/2ae6bdb60a46f22854c5ee965379bf0597bdb640/packages/Select/package.json#L36-L38

theetrain: My mistake, this should remain in `devDependencies` since it's only being used in tests.

PR #976, opened 2019-03-08
chore(other): add rebase on circle prepr step

marcod1419:
This will rebase when TDS Bot is being used to detect version changes. PR's commits exist on the local copy of master, and this script will pull origin/master and rebase onto the local copy of master.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -85,6 +85,23 @@ jobs:
     steps:
       - attach
       - expect
+      - add_ssh_keys:
+          fingerprints:
+            - '16:04:2d:ed:15:8d:0b:56:ea:ee:82:6e:01:a4:5b:09'
+      - run:
+          name: Add Github Key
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -85,6 +85,23 @@ jobs:
...
+      - run:
+          name: Configure Git Config & Branch
+          command: |
+            git status
+            git config --global user.email "tds.github.bot@gmail.com"
+            git config --global user.name "TDSBot"
+      - run:
+          name: Rebase branch
+          command: |
+            git checkout master
```
jraff: Can we just `git rebase master`? Wouldn't we have a clean master that we can rebase from?

marcod1419: Unfortunately no. It seems that circle ci actually adds commits from that branch to the top of its local master on its own, so rebasing won't work. Kind of odd behaviour but I can't find anything we did that does that so it may just be a normal thing.

PR #999, opened 2019-03-21
Add YoutubeVideo & VideoSplash

marcod1419:
This PR adds in `core-youtube-video` and `shared-video-splash`.

The former is a component that wraps the YouTube iframe api while enforcing some TDS sizing standards. Additionally, it works with `core-flex-grid`!

The latter is a splash screen that is shared by `core-youtube-video` and the upcoming `core-video`. This will give both the video component types some branding consistency.


packages/YoutubeVideo/YoutubeVideo.jsx
```diff
@@ -0,0 +1,107 @@
...
+}) => {
+  const [started, setStarted] = useState(false)
+
+  const aspectRatios = {
+    '16:9': { paddingTop: '56.25%' },
+    '4:3': { paddingTop: '75%' },
+    '1:1': { paddingTop: '100%' },
+  }
+
+  const AspectLimiter = styled.div({ ...aspectRatios[aspectRatio], position: 'relative' })
```
jraff: This should be moved outside of the render.

packages/YoutubeVideo/YoutubeVideo.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const initializeYoutubePlayer = event => {
```
jraff: This should also be moved outside of the render.

packages/YoutubeVideo/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "author": "TELUS digital",
+  "engines": {
+    "node": ">=8"
+  },
+  "bugs": {
+    "url": "https://github.com/telusdigital/tds-core/issues"
+  },
+  "homepage": "http://tds.telus.com",
+  "peerDependencies": {
+    "react": "^16.8.0",
```
jraff: Can we use "^16.8.2" to be inline with our other styled component refactors?

marcod1419: Yes, although I think we should be changing everything to 16.8.5 because of the newly added shallow rendering support on hooks right?

jraff: Wouldn't we only need to use `^16.8.5` in https://github.com/telus/tds-core/blob/master/package.json?

theetrain:
This was replaced with #1009

PR #1032, opened 2019-04-11
feat(core-box): allow custom styles through className & style prop

marcod1419:
This removes safeRest from Box, allowing styled components to be implemented more elegantly.

### DIFF
--- a/packages/Box/Box.jsx
+++ b/packages/Box/Box.jsx
@@ -1,9 +1,9 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
-import safeRest from '../../shared/utils/safeRest'
 import joinClassNames from '../../shared/utils/joinClassNames'
 import capitalize from '../../shared/utils/capitalize'
+import { deprecate } from '../../shared/utils/warn'
 
 import styles from './Box.modules.scss'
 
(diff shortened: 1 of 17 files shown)


packages/Box/Box.jsx
```diff
@@ -51,10 +50,11 @@ const Box = ({
     getClassName('padding', 'vertical', ySize),
     getClassName('margin', 'bottom', below),
     getBetweenClasses(between, inline),
-    dangerouslyAddClassName
+    dangerouslyAddClassName,
+    rest.className
```
jraff: `className` should be destructured in props

packages/Box/__tests__/Box.spec.jsx
```diff
@@ -90,11 +90,4 @@ describe('Box', () => {
 
     expect(box).toHaveClassName('a-class')
   })
-
-  it('does not allow custom CSS', () => {
```
jraff: Should we change this test to allow for `className` instead of removing it?

packages/Box/Box.jsx
marcod1419: That was done to avoid eslint demanding prop validation. Since it's a pretty standard react prop, should we have this?

packages/Box/__tests__/Box.spec.jsx
marcod1419: I think it's ok to not have that since that would be testing a React feature rather than something we did. That test was necessary because we actively blocked a React feature from working.

PR #1036, opened 2019-04-16
Convert Breadcrumbs to Styled Components

marcod1419:
This PR changes Breadcrumbs over to Styled Components. Additionally, it allows the use of `className` in `ColoredTextProvider`, as that is necessary for Styled Components. In the near future, we should probably refactor `ColoredTextProvider` or rethink its use entirely, as it is using a legacy React context API.


packages/Breadcrumbs/package.json
```diff
@@ -21,20 +21,21 @@
...
     "@tds/core-text": "^1.0.10",
+    "@tds/shared-typography": "^1.2.2",
     "@tds/util-prop-types": "^1.3.1",
     "prop-types": "^15.5.10"
   },
   "devDependencies": {
     "@tds/core-colours": "^1.0.0",
-    "@tds/core-responsive": "^1.2.0",
-    "@tds/shared-colored-text-provider": "^1.0.1"
+    "@tds/core-responsive": "^1.2.0"
```
theetrain: This needs to be a dependency since we're using its JS exports and not CSS.

packages/Breadcrumbs/Item/Item.jsx
```diff
@@ -1,14 +1,42 @@
...
+import styled from 'styled-components'
 
+import { colorTelusPurple, colorShuttleGrey } from '@tds/core-colours'
 import Link from '@tds/core-link'
+import { media } from '@tds/core-responsive'
 import Text from '@tds/core-text'
+import { sizeSmall, sizeMedium } from '@tds/shared-typography'
 
-import joinClassNames from '../../../shared/utils/joinClassNames'
 import ColoredTextProvider from '../../../shared/components/ColoredTextProvider/ColoredTextProvider'
```
theetrain: Update this to be a proper dependency, not a relative path

Issue #1044, opened 2019-04-23
OrderedList.Item renders incorrectly on Firefox and IE


rbrander:
## Description

When using a paragraph inside of OrderedList.Item, there will be a blank line before the content when viewed on IE and Firefox.

<img width="700" alt="Screen Shot 2019-04-09 at 2 18 03 PM" src="https://user-images.githubusercontent.com/968192/56590791-a5ffc700-65b5-11e9-81a0-06435301e20c.png">

NOTE: This only applies to OrderedList, as you can see the image renders the UnorderedList correctly.

## Reproduction Steps

```javascript
<OrderedList>
  <OrderedList.Item><Paragraph>First Item</Paragraph></OrderedList.Item>
  <OrderedList.Item><Paragraph>Second Item</Paragraph></OrderedList.Item>
  <OrderedList.Item><Paragraph>Third Item</Paragraph></OrderedList.Item>
</OrderedList>
```

1. Create an ordered list using the TDS core components
2. Embed a Paragraph (TDS core) component inside each OrderedList.Item
3. View rendered output in IE or Firefox

## Reproduction Percentage (if applicable)

- 10/10 times

## Meta

- TDS component version: @tds/core-ordered-list@2.0.16
- Willing to develop solution: No
- Has workaround: No
- High impact: No

## Screenshots

<img width="700" alt="Screen Shot 2019-04-09 at 2 18 03 PM" src="https://user-images.githubusercontent.com/968192/56590791-a5ffc700-65b5-11e9-81a0-06435301e20c.png">

pkandathil:
Seems to how FF and chrome render HTML https://codepen.io/anon/pen/dLQOxj

marcod1419:
@rbrander This issue has been corrected in `OrderedList` 2.0.20, and is also available in the styled-components version of `OrderedList`, 3.0.0. `UnorderedList` has also received the semantic changes. Thanks for reporting this!

PR #1049, opened 2019-04-26
 feat(core-unordered-list): added support for mixed lists

marcod1419:
This adds mixed icon support for UnorderedList via a new `itemStyle` prop on each list item. Additionally, it addresses the text issue from #1044 for UnorderedList, so it has parity when we apply the fix to OrderedList.

This should be merged before the Styled Components upgrade PR!


packages/UnorderedList/UnorderedItem/UnorderedItem.jsx
```diff
@@ -1,31 +1,43 @@
...
   </li>
 )
 
 UnorderedItem.propTypes = {
-  listStyle: PropTypes.oneOf(['circle', 'checkmark', 'x']),
+  /**
+   * The list's global bullet style. Can be overriden by listStyle.
+   * @ignore
+   */
+  listStyle: PropTypes.oneOf(['', 'circle', 'checkmark', 'x']),
```
theetrain: Having `''` as a prop value here is unintended. It should be `PropTypes.oneOf(['circle', 'checkmark', 'x']),`

packages/UnorderedList/UnorderedList.md
```diff
@@ -31,3 +31,18 @@
   <UnorderedList.Item>A mix of numbers, lowercase and uppercase letters</UnorderedList.Item>
 </UnorderedList>
 ```
+
+### Mixed List
+
+If you do not want each list item to have the same icon, you may give each item its own icon via the `itemStyle` prop. This will override a `listStyle` set by the parent.
```
theetrain: Note here, or in the general Lists.md file, usage criteria for mixed list styles:

- Mixed lists should only be used within Notification or InputFeedback components

marcod1419:
Note that this needs acceptance criteria before merging! Otherwise, this can be QA'd.

PR #1084, opened 2019-05-15
fix(shared-choice): show full label width when error is active on ie 11

marcod1419:
This fixes a bug reported by @coltonpowershifter where input labels may not display properly in IE 11 if an error is active. (This issue is already fixed in the styled-components PR of checkbox and radio)


theetrain:
QA looks good

PR #1096, opened 2019-05-30
feat(core-select): convert to styled-components

marcod1419:
This converts `Select` to styled-components. Additionally, the following fixes were implemented:

- `value` will now override `defaultValue`
- error messages can only appear when `feedback` is set to 'error'
- improved documentation


packages/Select/Select.jsx
```diff
@@ -1,70 +1,267 @@
-import React from 'react'
+import React, { useState } from 'react'
 import PropTypes from 'prop-types'
-import { componentWithName } from '@tds/util-prop-types'
+import styled from 'styled-components'
 
-import DecorativeIcon from '@tds/core-decorative-icon'
 import Box from '@tds/core-box'
+import DecorativeIcon from '@tds/core-decorative-icon'
+import { FeedbackIcon } from '@tds/shared-form-field'
```
jraff: Should we be using `FeedbackIcon` from `@tds/core-input`?

PR #1097, opened 2019-05-30
fix(core-video): ui sizing changes

marcod1419:
- side padding reduced when player is small
- replay button made smaller
- hours number removed, leading 0 for minutes removed
- volume slider bigger when player is small
- prevent an issue where seeking would misbehave with multiple videos on the same page


packages/Video/ControlBar/Controls/VideoProgressBar/VideoProgressBar.jsx
```diff
@@ -4,6 +4,8 @@ import styled from 'styled-components'
 
 import { colorTelusGreen } from '@tds/core-colours'
 
+const videoProgressBar = React.createRef()
```
theetrain: 👌

PR #1101, opened 2019-06-05
feat(core-text-area): upgrade to styled-components

marcod1419:
Converts `TextArea` to styled-components and improves its docs.


packages/TextArea/TextArea.jsx
```diff
@@ -1,40 +1,212 @@
...
+      </Box>
+    )
+  }
 )
 
+TextArea.displayName = 'TextArea'
+
 TextArea.propTypes = {
+  /**
+   * The id. If no `id` is provided, a default `id` will be generated in the format of: "`name`\_`value`".
```
theetrain: Please add a clearer meaning of the generated ID's output, using an example.

packages/TextArea/TextArea.jsx
```diff
@@ -54,7 +226,7 @@ TextArea.propTypes = {
   /**
    * An error message. Either an error or a helper should be used, not both.
    */
-  error: PropTypes.string,
+  error: PropTypes.node,
```
theetrain: Why was this changed to node?

packages/TextArea/package.json
```diff
@@ -20,20 +20,25 @@
...
     "prop-types": "^15.5.10"
   },
   "devDependencies": {
-    "@tds/core-colours": "^1.1.2",
-    "@tds/core-input-feedback": "^1.0.13",
-    "@tds/core-paragraph": "^1.0.12",
-    "@tds/core-standalone-icon": "^1.1.0",
-    "@tds/core-text": "^1.1.0",
-    "@tds/shared-animation": "^1.2.0"
+    "react-transition-group": "^2.2.1"
```
theetrain: Is this being used anywhere?

packages/TextArea/TextArea.jsx
```diff
@@ -1,40 +1,212 @@
...
+  colorShuttleGrey,
+  colorWhite,
+  colorPrimary,
+  colorCardinal,
+  colorAthensGrey,
+} from '@tds/core-colours'
+import { medium, mediumFont, color } from '@tds/shared-typography'
+import { FeedbackIcon } from '@tds/core-input'
+
+import safeRest from '../../shared/utils/safeRest'
```
theetrain: Please add this to `devDependencies`

packages/TextArea/TextArea.jsx
marcod1419: It's possible to pass in tds links now.

packages/TextArea/TextArea.jsx
marcod1419: Confirmed that this is not a package in core.

PR #1112, opened 2019-06-11
Add simpleMode and videoBorder options

marcod1419:
Due to internal requirements, a simple video player mode has to be added as a feature. Additionally, a new border prop has been added for said simple player mode.

This is a draft because I'm going to add a warning real quick when the border prop is used without simple mode being enabled.


packages/Video/Video.jsx
```diff
@@ -347,6 +360,12 @@ class Video extends React.Component {
 
   // ******** Begin Helper Functions *********
 
+  showWarnings = () => {
```
jraff: This is not needed. The `videoBorder` prop can be used as long as the video has a white background.

packages/Video/Video.jsx
```diff
@@ -798,6 +824,17 @@ Video.propTypes = {
...
    */
   crossOrigin: PropTypes.oneOf(['anonymous', 'use-credentials']),
+  /**
+   * Disables the control bar during playback and reduces the amount of time until the UI hides itself.
+   * For special approved internal uses only.
+   * @since 1.2.0
+   */
+  simpleMode: PropTypes.bool,
+  /**
+   * Renders a gainsboro coloured border around the video. Used to seperate the video from the rest of the page if the video's background is the same colour as the container's. Only to be used with `simpleMode` enabled.
```
jraff: Can be used in both `simpleMode` and normal mode.

PR #1114, opened 2019-06-14
feat(core-card): upgrade to styled components

marcod1419:
BREAKING CHANGE: styled-components now a peer dependency


packages/Card/Card.jsx
```diff
@@ -1,20 +1,35 @@
...
+import { borders } from '@tds/shared-styles'
 import safeRest from '../../shared/utils/safeRest'
 
-import styles from './Card.modules.scss'
-
+const StyledCard = styled(({ variant, ...rest }) => <Box {...rest} />)(({ variant }) => ({
+  ...borders.none,
+  ...borders.rounded,
+  ...(variant === 'white' && {
+    boxShadow: '0 0 16px 0 rgba(213, 213, 213, 0.5)',
```
theetrain: Is there an opportunity to move this to `@tds/shared-styles`?

packages/Card/__tests__/Card.spec.jsx
```diff
@@ -14,16 +14,16 @@ describe('<Card />', () => {
 
   it('can be presented as one of the allowed variants', () => {
     let card = doShallow()
-    expect(card.dive()).toHaveClassName('white')
+    expect(card.dive()).toMatchSnapshot()
```
theetrain: These need to be mounted in order for the snapshot to contain rendered CSS.

packages/Card/Card.jsx
marcod1419: I don't think so, this is only used in Card as far as I'm aware.

Issue #1123, opened 2019-06-19
Input (type=email) cursor moves when prefixed with whitespace


rbrander:
## Description

I'm trying to use an Input component to capture email addresses.  When the type is set to 'email', the behaviour of the input changes (as compared to type='text'), specifically when the field value is prefaced with whitespace.  Once a character is added after the whitespace, the cursor moves to the beginning of the field and the value is changed to have only the new character (without whitespace).

## Reproduction Steps

1. Render an Input component with type="email"
2. Click on the Input component and type one or many spaces
3. Type 'a', 's'
4. Observe the field value is 'sa', with the cursor between the two characters

I expected to see 'as' with the cursor at the right of the 's'

## Reproduction Percentage (if applicable)

- 10/10 times

## Meta

- TDS component version: @tds/core-input@3.0.2
- Willing to develop solution: No
- Has workaround: Yes, use a type="text"
- High impact: Yes, Form Builder requires an email specific field prior to launch

jraff:
To add to this issue, it looks to only be an issue on WebKit based browsers (Chrome, Safari).

agorovyi:
The fix has been released: https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-input%403.0.3

PR #1127, opened 2019-06-19
fix(core-input): remove value from props when input is uncontrolled

jraff:
## Related issues

See #1123 

## Description

Do not include `value` prop if using an uncontrolled `Input`.

- [x] New code is unit tested

### DIFF
--- a/packages/Input/__tests__/Input.spec.jsx
+++ b/packages/Input/__tests__/Input.spec.jsx
@@ -122,22 +122,17 @@ describe('Input', () => {
 
   describe('editability', () => {
     it('supports string values or number values', () => {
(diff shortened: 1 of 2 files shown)


packages/Input/__tests__/Input.spec.jsx
```diff
@@ -122,24 +122,19 @@ describe('Input', () => {
 
   describe('editability', () => {
     it('supports string values or number values', () => {
-      let findInputElement = doMount().findInputElement
-      expect(findInputElement()).toHaveValue(undefined)
+      const onChangeMock = jest.fn()
```
marcod1419: Good call.

Issue #1156, opened 2019-06-26
Improve `withFocusTrap` using aria attributes


cuginoAle:
## Description
When showing modal-dialogs (i.e. the mobile burger-menu content, the channels page filters popup on mobile...)  the screen-reader virtual cursor can easily escape the focus trap.

## Recommendation
Enhance the `withFocusTrap` module to make use of `aria-modal='true'` and `role='dialog'`. 
In my experience it is important to set also `aria-label` to ensure the modal mode is triggered and **set the focus inside the modal as soon as the content loads.**

Those settings have been successfully tested on Chrome/Safari/Brave/Opera/Firefox on OSX and Chrome/Firefox on Windows using both **NVDA** and **VoiceOver**.

The only main browser not implementing `aria-modal` properly is **Ms Edge**, as it allows the user to escape the focus trap.

Also, the `withFocusTrap` component is currently coded as a **HOC**. Since it appears to be just a wrapper (it doesn't provide any **derived prop** or any **callback**) it would make more sense to use the Wrapper Component technique to promote an easier and more declarative use of the component.
In essence you would use it like so:
```html
<FocusTrap>
    <MyDialogContent>
     ...
   </MyDialogContent>
</FocusTrap>

```

- TDS component: `tds-core/packages/hocs/withFocusTrap.jsx`
- Willing to develop solution: Yes
- Has workaround: Yes/No 
- High impact: No

PR #1157, opened 2019-06-27
feat(packages): withFocusTrap aria attributes

cuginoAle:
## Related issues

fixes: #1156 

## Description
Adding `aria` attributes to support modal-dialogs behaviour for ATs.
Makes use of `useEffect` to set the focus inside the dialog on render

- [x] New code is unit tested


packages/hocs/withFocusTrap.jsx
```diff
@@ -1,27 +1,47 @@
...
 
 const withFocusTrap = Component => {
   const WithFocusTrap = props => {
     const componentRef = useRef(null)
 
+    useEffect(() => {
+      // setting the focus to the first focusable element
+      const focusableElements = getFocusable(componentRef.current)[0]
+      if (focusableElements) focusableElements.focus()
+    })
```
jraff: This should only run on mount. Please add `[]` as the second parameter to `useEffect`

packages/hocs/withFocusTrap.jsx
```diff
@@ -1,27 +1,47 @@
 /* eslint-disable jsx-a11y/no-noninteractive-tabindex */
-import React, { useRef } from 'react'
+import React, { useRef, useEffect } from 'react'
+import PropTypes from 'prop-types'
 
 // Returns an array of focusable elements in the order they are found in c
-const getFocusable = c =>
-  c.querySelectorAll('button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])')
+const selector =
+  'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"]), audio[controls], video[controls], [contenteditable]:not([contenteditable=false]'
```
jraff: Can we remove `:not([tabindex="-1"])`. Example, we need to focus on a `tabIndex={-1}` in our [Footnote](https://github.com/telus/tds-core/blob/master/packages/TermsAndConditions/Footnote/Footnote.jsx#L268) component

packages/hocs/withFocusTrap.jsx
```diff
@@ -1,27 +1,47 @@
...
+      <div role="dialog" aria-modal="true" aria-label={ariaLabel}>
         <div onFocus={handleFocus(false)} tabIndex={0} />
-        <Component {...props} ref={componentRef} />
+        <Component {...rest} ref={componentRef} />
         <div onFocus={handleFocus(true)} tabIndex={0} />
       </div>
     )
   }
+  WithFocusTrap.propTypes = {
+    ariaLabel: PropTypes.string,
```
jraff: Can we rename this prop to `a11yText` to be consistent with the rest of the design system?

packages/hocs/withFocusTrap.jsx
```diff
@@ -1,27 +1,47 @@
...
       const focusableElements = getFocusable(componentRef.current)
       focusableElements[isEnd ? 0 : focusableElements.length - 1].focus()
     }
 
     return (
-      <div>
+      <div role="dialog" aria-modal="true" aria-label={ariaLabel}>
         <div onFocus={handleFocus(false)} tabIndex={0} />
-        <Component {...props} ref={componentRef} />
+        <Component {...rest} ref={componentRef} />
```
jraff: This assumes the `Component` can accept a ref. Can you add a code comment that informs users that the `Component` must be able to accept a ref.

packages/hocs/withFocusTrap.jsx
cuginoAle: 👍

packages/hocs/withFocusTrap.jsx
jraff: Although W3C says that the focus should be set to the first interact-able element (which in many cases is a "close" button). We need to look at what the user experience is. In the case of the TDS [Footnote](https://tds.telus.com/components/index.html#!/Footnote/1), we have a `tabIndex` of -1 on title, we want focus set there immediately so that the user knows where they are. Once focus has been moved away from the title, we do not want it keyboard focusable.

@cuginoAle What are your thoughts?

jraff: @cuginoAle and I have decided that we will keep `[tabindex]:not([tabindex="-1"]` and allow for an opt-out of `withFocusTrap` setting initial focus.

cuginoAle:
- Adding support for `autofocus` property
- setting the `ref` attribute  to the wrapping `div` to allow `<Component>` to be a stateless component and to make sure `componentRef.current` is always an html node

PR #1173, opened 2019-07-08
TermsAndConditions fixes

jraff:
### TermsAndConditions

* Line-height of the chevron is different is affected by resets.
* Parse anchor tags in case someone adds them as HTML to legal copy
* Add tabIndex to heading and focus on `Footnote` open

### Heading

* Add forwardRefs

- [ ] New code is unit tested

### DIFF
--- a/packages/TermsAndConditions/Footnote/Footnote.jsx
+++ b/packages/TermsAndConditions/Footnote/Footnote.jsx
@@ -265,7 +265,7 @@ const Footnote = props => {
                 <FlexGrid.Row>
                   <FlexGrid.Col xs={12}>
                     <StyledHeader between="space-between" inline>
-                      <Heading level="h4" tag="h2" tabIndex={0} ref={headingRef}>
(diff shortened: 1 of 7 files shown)


packages/TermsAndConditions/Footnote/Footnote.jsx
```diff
@@ -240,7 +241,7 @@ const Footnote = props => {
   // focus on the close button on open/content change
   useEffect(() => {
     if (content !== null && isVisible && closeRef && closeRef.current !== null) {
```
agorovyi: Should the condition also check for `headingRef` instead of `closeRef` since you changed it?

```
(content !== null && isVisible && headingRef && headingRef.current !== null)
```

jraff: Yes, in fact we don't need `closeRef` at all anymore. I've removed it completely.

PR #1175, opened 2019-07-09
Upgrade Box to Styled Components

marcod1419:
This upgrades `Box` to styled components while also removing the `dangerouslyAddClassname` prop.


packages/Box/Box.jsx
```diff
@@ -1,70 +1,113 @@
...
+      marginBottom: space,
+    }),
+    size
   )
 }
 
+const StyledBox = styled.div.attrs(({ className, tag }) => {
+  return { className, as: tag }
+})(({ inline, between, horizontal, vertical, inset, below }) => ({
+  display: between ? 'flex' : 'block',
```
jraff: Instead of spreading, we should pass each function as an argument to styled(). This way, 'styled-components' will deeply merge all properties.

packages/Box/Box.jsx
```diff
@@ -98,23 +141,12 @@ Box.propTypes = {
    *
    * By default, `between` will arrange the Box's children as a flex column. Combine with `inline` to arrange them
    * as a flex row.
-   *
-   * Use `space-between` to set an equal amount of space between all items, within the bounds of the parent.
```
jraff: This should be included in the styled box.

PR #1186, opened 2019-07-16
feat(core-step-tracker): updgrade to styled-components, add copy prop

marcod1419:
BREAKING CHANGE: styled-components is now a peer dependency

This also adds the `copy` prop and deprecates the old `mobileStepLabelTemplate` prop. In the next major release, the `copy` prop should be required.


packages/StepTracker/__tests__/StepTracker.spec.jsx
```diff
@@ -37,35 +38,24 @@ describe('<StepTracker />', () => {
     expect(stepTracker.find('[data-testid="step-0"]')).toContainReact(
       <Icon symbol="checkmark" size={16} variant="inverted" />
     )
-    expect(stepTracker.find('[data-testid="singleStepContainer-0"]')).toHaveProp(
-      'data-isactive',
-      true
-    )
+
+    expect(stepTracker).toMatchSnapshot()
```
agorovyi: Is there a way to test this other way than snapshots? All these snapshot matchers made the actual snapshot file too big which will make it harder to read later when it breaks.

packages/StepTracker/package-lock.json
```diff
@@ -0,0 +1,41 @@
+{
```
agorovyi: This lock file should not be committed I think.

agorovyi:
I think the branch also needs a rebase.

Issue #1131, opened 2019-06-20
[core-select] add 'hintPosition' and allow 'below' as option


Nintendot:
## Description

Current [core-select] has prop `hint` but it's position is fixed as `inline`, we would like to have `below` as an option, similar to how [core-input] handles `hint`

## Recommendation

Enable `hintPosition` and have `below` as an option

## Designs

n/a

## Meta

- TDS component: [core-select] 
- Willing to develop solution: Yes/No
- Has workaround: Yes/No 
- High impact: Yes/No 
- Designer/Developer Pair: Your name, your design/dev pair's name

Nintendot:
cc: @rbrander

agorovyi:
This is done and [released in v3.1.0](https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-select%403.1.0)

mejsullivan:
Does this "hint below" associate with the input field properly to provide an error notification to the screen reader user and keyboard user at the correct times?

PR #1188, opened 2019-07-17
feat(core-select): add hintPosition below

jraff:
## Related issues

See #1131 

- [ ] New code is unit tested


packages/Select/Select.jsx
```diff
@@ -131,18 +131,21 @@ const renderError = (error, errorId) => (
...
-        <Text size="medium" bold data-testid="selectLabel">
-          {label}
-        </Text>
-        {hint && renderHint(hint, Text, hintId)}
-      </StyledLabelContainer>
-    </label>
-    {tooltip && React.cloneElement(tooltip, { connectedFieldLabel: label })}
-  </Box>
+const renderLabel = (id, label, hint, hintPosition, hintId, tooltip) => (
+  <div>
```
agorovyi: Can this `<div>` be replaced with Fragment?

packages/Select/Select.jsx
```diff
@@ -279,6 +283,10 @@ Select.propTypes = {
    * Clarify the expected input.
    */
   hint: PropTypes.string,
+  /**
+   * Position of the `hint` relative to `label`.
+   */
```
agorovyi: Could you also add `@since x.x.x` and specify which version this prop is available since?

jraff: Added.

packages/Select/Select.jsx
jraff: No, it needs to use the `div` so as to not be affected by the `Box`

PR #1195, opened 2019-07-23
Upgrade FlexGrid to Styled Components, Export Breakpoints From Responsive

marcod1419:
This upgrades our FlexGrid override CSS to styled-components, while also upgrading the dependency FlexGrid relies on for a slightly cleaner DOM.

To facilitate FlexGrid, Responsive has been updated to export an object containing our breakpoints.

Lastly, I went ahead and ran `npm audit` to clear up lots of vulnerabilities.


packages/FlexGrid/__tests__/FlexGrid.spec.jsx
```diff
@@ -109,6 +104,6 @@ describe('FlexGrid', () => {
       limitWidth: false,
     })
 
-    expect(flexGrid).not.toHaveClassName('limitWidth')
+    expect(flexGrid).toMatchSnapshot()
```
agorovyi: Adding too many snapshot assertions is stretching the final snapshot file which will make it complicated to read when it breaks in future. It's a good practice to keep snapshot files short.

marcod1419: How do you recommend we address this? This was checking for class names originally, and the main recommendation when using styled-components is to use snapshot testing.

agorovyi: What do you think of this refactor? Test for line 52:

```
  ...
  findStyledColumn: index => flexGrid.find(`StyledComponent#col-${index}`),
  ...

  it('should render all children related to Grid with no gutter', () => {
    const { findStyledColumn, findRow } = doMount({ gutter: false })

    const styledCol1 = findStyledColumn(1)
    const styledCol2 = findStyledColumn(2)
    const row = findRow(1)

    expect(styledCol1.prop('gutter')).toEqual(false)
    expect(styledCol2.prop('gutter')).toEqual(false)
    expect(row).toHaveClassName('row')
  })
```

agorovyi:
I rebased from master to get a new contributor commit instead of committing it again.

PR #1205, opened 2019-07-31
chore(scripts): exit e2e docker container when finished

marcod1419:
Fixes the issue where the e2e server will stay open after tests have completed.


theetrain:
This is no longer needed since we're moving away from Docker.

Issue #1233, opened 2019-08-14
Some text is squished in app due to letter-spacing


msvivianso:
## Description

The `letter-spacing: -0.8px;` styling is affecting the text in the app to look squished. We inspected the affect elements and it looks like this styling is coming from `@tds/core-css-reset`

After talking with @agorovyi, he was able to reproduce the issue in the starter kit. He mentioned that it's happening due to `<span>` being inside `Heading` component.

Reference to slack thread about this issue: https://telusdigital.slack.com/archives/C66LP1WGG/p1565802218147600

## Reproduction Steps

1. Add a `<span>` div with a text inside the `Heading` component
2. Load a page with the above styling
3. See squished text inside the span

This styling is also in the `Text` component.

## Meta

- TDS component version: @tds/core-heading
- Willing to develop solution: No
- High impact: No

## Screenshots

**Expected** (`letter-spacing: 0px;`)
<img width="1680" alt="Screen Shot 2019-08-14 at 11 00 09 AM" src="https://user-images.githubusercontent.com/6042193/63055301-97665400-beb3-11e9-86ef-5b1129d8b0dd.png">

**Actual** (`letter-spacing: -0.8px;`)
<img width="1680" alt="Screen Shot 2019-08-14 at 11 00 02 AM" src="https://user-images.githubusercontent.com/6042193/63055337-a947f700-beb3-11e9-9431-38caaca2d3a0.png">

theetrain:
Thanks for reporting. We're looking into a couple potential solutions:

### 1) Apply `letterSpacing: inherit` to `span`s within Heading and Typography components

By adding this selector to core-heading and other components, we can fix the edge case of nested `span`s not inheriting the `heading`'s letter spacing.

**Heading.jsx**
```diff
  h1: {
    ...helveticaNeueLight45,
    fontSize: '1.75rem',
    lineHeight: '1.29', // 36px
    letterSpacing: '-1.6px',
    ...media.from('md').css({
      ...helveticaNeueThin35,
      fontSize: '2.75rem',
      lineHeight: '1.18',
      letterSpacing: '0',
    }),
    [`${StyledHeadingSup}, sup`]: {
      ...baseSupSubScripts,
      fontSize: '1.25rem',
      top: '-1em',
      ...media.from('md').css({ fontSize: '1.25rem', top: '-1.3em' }),
    },
+   '& > span': { letterSpacing: 'inherit' }
  },
```

This solution **can also be used as a workaround** for Heading:

```jsx
[...]

PR #1235, opened 2019-08-15
fix(core-heading): add letterSpacing inherit to Heading child

agorovyi:
## Related issues

See #1233 

- [ ] New code is unit tested


packages/Heading/Heading.jsx
```diff
@@ -81,6 +81,9 @@ export const StyledHeading = styled.h1(wordBreak, ({ level, invert }) => {
   return {
     color,
     ...HeadingLevels[`${level}`],
+    '& > *': {
```
theetrain: Should we target `span` specifically? I wonder if this selector will affect superscripts.

agorovyi: I couldn't think of other scenarios but `span` but thought there could be some edge-cases. Since most of other edge-cases will go against accessibility (like nested `div` for example) I think we should target `span` specifically.

agorovyi: This is done.

Issue #907, opened 2019-01-28
TDS Core Spinner Component does not prevent updating of fields while spinning


akdambenieks:
## Description

TDS Core `Spinner` component:
While spinner is spinning it is still possible to keyboard tab into input fields to update values and buttons to resubmit events.

## Reproduction Steps

Enter input into a field and submit the form.  The spinner component wrapping the inputs receives the spinning attribute while a call to a rather slow API is made.  While the spinner is still 'spinning' it is possible to tab/shift+tab into the input fields to change their values or to tab onto the submit button and resubmit the form.  The spinner does however prevent mouse clicking into the wrapped inputs/buttons.

## Reproduction Percentage (if applicable)

Reproducible at least 9/9 times.  Others on my team as well as from DRB have no problem reproducing the issue.

## Meta

- TDS version: , Spinner version: 2.1.1
- Willing to develop solution: No
- Has workaround: Is possible
- High impact: Potentially yes, especially so for visually impaired users or for prolonged spinning events where repeated submission of forms can hang up the application.

## Screenshots

CODE USED TO IMPLEMENT THE SPINNER COMPONENT:
```jsx
    <FlexGrid limitWidth={false} gutter={false}>
        <Spinner inline spinning={isLoading || isLoadingAppointments} a11yLabel="Currently searching, please wait for results to be returned" tip="Working..">
          <FlexGrid.Row>
            <FlexGrid.Col>
              <Box between={4}>
                <Box between={3}>
                  {validationStatus === 'ERROR' &&
                    <Notification variant="error">
                      Please search and select an address. Orders cannot be created without a valid service address.
                    </Notification>
                  }
                  <Input
                    id="customerAddress"
                    value={address}
                    onChange={this._updateInput}
                    label="Street address"
                    aria-label="Customer's street address"
                    helper={AddressFormatHelper}
                  />
                  <Input
                    id="customerAddressInit"
                    value={addressUnit}
                    onChange={this._updateAddressUnit}
                    label="Unit number"
                    aria-label="Customer's unit number if applicable"
                    hint="if applicable"
                  />
                </Box>
                <Box>
                  <Button onClick={this._sendAutocomplete}>Search<A11yContent>for matching service address. Once results are returned please select one of the available addresses below</A11yContent></Button>
                </Box>
                <span aria-live="polite">
                  {addressStatus === 'NOMATCH' ?
                      'No results found for this address' : ''
                  }
                </span>
                {autocompleteList.length > 0 &&
[...]

PR #1247, opened 2019-08-27
Spinner prevent updating

jraff:
## Related issues

See #907 

## Description

- Add `labelRef` property to Spinner
- Add `tabIndex="-1"` to all children while Spinner is active

- [x] New code is unit tested


packages/Spinner/Spinner.jsx
```diff
@@ -43,6 +43,20 @@ const OpaqueContainer = styled.div({
   opacity: 0.06,
 })
 
+const recursiveMap = (children, fn) =>
```
theetrain: This could work, but I'd like to have this tested with 20 child nodes first. 

- Create a form with 20 core-input components, and two core-heading components
- Toggle and untoggle the spinner with a button, see if there is noticeable jank - test with dev tools running at 6x CPU slowdown
- Also set `inert={true}` to each child to provide a progressive enhancement for touch devices

packages/Spinner/Spinner.jsx
```diff
@@ -154,7 +171,7 @@ class Spinner extends React.PureComponent {
               e.target.blur()
```
theetrain: I think this wrapper element cannot trigger `onFocus` because it isn't a focusable element itself. Set `tabindex={0}` so that it becomes focusable. Perhaps if this solution works, we may not need to recursively set attributes to all children.

jraff: Focus events from any children will bubble up to this element, that being said. It doesn't really look like this is doing anything at all.

Issue #1287, opened 2019-10-16
[WebVideo] provide onReady callback


hamedmam:
## Add ability to pass down analytics events trigger to the WebVideo component onReady 

- Home solution team is looking to add analytics events to the WebVideo that will get triggered once the WebVideo is ready to play. below screenshot is how they implemented in the fork version of Site Builder's BlockVideo. this component was forked due to the issue with React hooks on Site Builder and now that issue is resolved we would like to proceed to replace the forked code with WebVideo component therefore we need to find a way to pass down analytics events.

## Recommendation

- enable the WebVideo component to pass down a function that gets triggered onReady

## Designs

N/A

## Meta

- TDS component: @tds/core-web-video
- Willing to develop solution: No
- Has workaround: No
- High impact: Yes
- Designer/Developer Pair: please reach out to myself or @mike-bunce for more information

<img width="514" alt="Screen Shot 2019-10-16 at 10 43 06 AM" src="https://user-images.githubusercontent.com/24867760/66930518-b80d7000-f002-11e9-804c-6955a43749f8.png">

jraff:
Fixed with [@tds/core-web-video@1.0.1](https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-web-video%401.1.0)

@hamedmam @mike-bunce @nikaukraine

PR #1291, opened 2019-10-17
feat(core-web-video): add onstart callback

marcod1419:
See #1287 

This adds an `onStart` callback to `WebVideo`. It's called `onStart` instead of `onReady` because the function gets called when play is pressed on the video splash screen, rather than when the component is loaded.

### DIFF
--- a/packages/WebVideo/WebVideo.jsx
+++ b/packages/WebVideo/WebVideo.jsx
@@ -46,11 +46,13 @@ const WebVideo = ({
   beginMuted,
   videoLength,
   copy,
+  onStart,
   ...rest
 }) => {
   const [started, setStarted] = useState(false)
 
   const initializeYoutubePlayer = event => {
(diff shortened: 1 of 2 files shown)


packages/WebVideo/WebVideo.jsx
```diff
@@ -46,11 +46,13 @@ const WebVideo = ({
   beginMuted,
   videoLength,
   copy,
+  onStart,
   ...rest
 }) => {
   const [started, setStarted] = useState(false)
 
   const initializeYoutubePlayer = event => {
+    onStart()
```
theetrain: If `onStart` is `undefined`, I'm not sure if this will throw an error.

marcod1419: It doesn't, but I changed the default to `()=>{}` anyway since it's more proper.

PR #1307, opened 2019-11-01
Fix fade animation on FadeAndReveal

marcod1419:
Quick PR to patch the broken fade out animation on the shared FadeAndReveal component.

### DIFF
--- a/packages/Animation/FadeAndReveal.jsx
+++ b/packages/Animation/FadeAndReveal.jsx
@@ -30,7 +30,7 @@ const transitionStyles = (height, timeout) => ({
     opacity: 1,
     height,
     visibility: 'visible',
-    transition: 'unset',
+    transition: `height ${timeout}ms ease-in-out, opacity ${timeout}ms ease-in-out, visibility 0ms ${timeout}ms`,
   },
   exited: {
     opacity: 0,
--- a/packages/Animation/FadeAndReveal.jsx
+++ b/packages/Animation/FadeAndReveal.jsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { Transition } from 'react-transition-group'
+import { CSSTransition } from 'react-transition-group'
 import styled from 'styled-components'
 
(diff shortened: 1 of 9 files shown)


packages/Animation/FadeAndReveal.jsx
```diff
@@ -8,31 +8,40 @@ export const StyledContainer = styled.div({
...
 
 const transitionStyles = height => (duration, timeout) => ({
-  entering: { opacity: 1, height },
-  entered: { opacity: 1, height: 'auto' },
+  entering: {
+    opacity: 1,
+    height,
+    visibility: 'visible',
+    transition: `height ${duration || timeout}ms, opacity ${duration ||
+      timeout}ms ease-in-out, visibility 0s`,
```
theetrain: Can you transition `visibility`?

jraff: I don't think we need `visibility 0s` here.

packages/Animation/FadeAndReveal.jsx
```diff
@@ -8,31 +8,40 @@ export const StyledContainer = styled.div({
...
     opacity: 0,
     height,
-    transition: `height ${duration || timeout}ms, opacity ${duration || timeout}ms ease-in-out`,
   },
-  exited: { opacity: 0 },
+  exited: {
+    opacity: 0,
+    visibility: 'hidden',
+    transition: `height ${duration || timeout}ms, opacity ${duration ||
+      timeout}ms ease-in-out, visibility 0s ${(duration || timeout) / 2}ms`,
```
jraff: I don't think we need visibility 0s here.

marcod1419: Visibility here is necessary because it leverages a delay, which causes it to kick in after the panel has safely collapsed.

packages/Animation/FadeAndReveal.jsx
marcod1419: Yes and no. In this case, we use a delay, which does work.

Issue #1320, opened 2019-11-18
[TermsAndConditions] no expanded announcement to assistive technologies


theetrain:
## Description

Originally reported by @owestinTelus 

The TermsAndConditions component doesn't announce its expanded or collapsed state to assistive technologies.

## Reproduction Steps

1. Use Voiceover on Safari while observing the TermsAndConditions component
2. Expand the T&C
3. Observe there is not 'expanded' announcement, as there is in ExpandCollapse

## Recommendation

- Provide `aria-expanded` attribute to T&C, and update it based on its expanded state

## Meta

- TDS component version: `@tds/core-terms-and-conditions@1.2.2`
- Willing to develop solution: No
- High impact: No

marcod1419:
This has been resolved in the latest TermsAndConditions update: https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-terms-and-conditions%401.2.6

Thanks for reporting @owestinTelus!

PR #1323, opened 2019-11-20
fix(core-terms-and-conditions): add aria-expanded to TermsAndConditons

hamedmam:
## Related issues
https://github.com/telus/tds-core/issues/1320

## Description
Adds aria-expanded attribute to the button on Terms and Condition to enable screen reader announce the expanse-collapse states

- [x] New code is unit tested


jraff:
Please change this commit type to a `fix`.

PR #1327, opened 2019-11-27
fix(core-terms-and-conditions): prevent chevron icon from shifting

marcod1419:
Quick PR to fix an issue I noticed in Terms And Conditions. In desktop, the chevron expand/collapse icon will shift slightly when opened and closed. This addresses that.


marcod1419:
Gotta fix something, will re-create!

PR #1338, opened 2019-12-16
Junk/test sauce

theetrain:
This should fail. plz don't merge unless you want pink buttons in the design system.


theetrain:
No thank you, we're not looking into changing the branded appearance of core-button at this time. Thanks for your contribution!

PR #1341, opened 2019-12-17
fix(core-interactive-icon): change invert cart filled dot to telus green

marcod1419:
This PR changes the colour of the green dot on the cart filled nav icon to improve colour contrast on dark backgrounds.


packages/InteractiveIcon/svgs/NavButton/CartFilledBold.jsx
```diff
@@ -45,6 +46,7 @@ CartFilledBold.propTypes = {
       a11yText: PropTypes.string.isRequired,
     }),
   ]).isRequired,
+  variant: PropTypes.oneOf(['default', 'alternative', 'inverted']).isRequired,
```
jraff: Why do we have alternative here? NavButton does not have an alternative variant.

PR #1342, opened 2019-12-18
fix(core-interactive-icon): improve invert navbutton focus/active style

marcod1419:
This PR improves the focus and active styles on nav buttons when being inverted.


packages/InteractiveIcon/components/Tooltip.jsx
```diff
@@ -16,11 +16,11 @@ const StyledTooltip = styled.div(
     padding: '0.0625rem 0.5rem 0.1875rem 0.5rem',
     maxWidth: '8.25rem',
     backgroundColor: colorGreyShuttle,
-    border: `1px solid ${colorWhite}`,
```
jraff: The tooltip should still have a 1px border.

packages/InteractiveIcon/shared/StyledInteractiveIconButton.jsx
```diff
@@ -15,7 +15,13 @@ const getOutline = ({ variant }) => {
...
   return {
     '&:focus': {
-      outline: `0.125rem solid ${colorWhite}`,
+      outline: 'transparent',
+      border: `0.125rem solid ${colorWhite}`,
+      borderRadius: '50%',
+    },
+    '&:active': {
+      backgroundColor: 'rgba(0,0,0,0.5)',
+      backgroundBlendMode: 'multiply',
```
jraff: `background-blend-mode` is not supported in IE11 or Edge. Will need to keep an eye on this during QA.

PR #1352, opened 2020-01-08
fix(core-notification): throw console warn on dismissable error/warning

marcod1419:
This PR will cause a console warning to be thrown when `dismissible` is used in combination with the `error` or `warning` variants. Additionally, it improves the overall documentation of Notification.


packages/Notification/Notification.md
```diff
@@ -68,7 +76,14 @@ Use the `warning` variant to provide feedback of a warning; users would still be
...
 ### Dismissible
 
-Use the `dismissible` prop to allow users to dismiss the `Notification` at any time. The `dismissible` prop must be used together with the `dismissibleA11yLabel` to remain accessible. Once dismissed, the `Notification` is removed from the DOM.
+Notifications can either be passive or require action from the user. If the information is passive then it should be dismissible as indicated by an interactive close button on the far right. If the Notification requires an action then make the option(s) clear with a Chevron Link or Paragraph link style.
+
+Use the dismissible prop to allow users to dismiss the Notification at any time. The `dismissible` prop must be used together with the `dismissibleA11yLabel` to remain accessible. Once dismissed, the Notification is removed from the DOM.
+
+Please note:
+
+- Branded, General and Success and Instructional variants can be dismissed
```
jraff: Lets remove **"and Instructional"**, why are we capitalizing General, and Success?

packages/Notification/Notification.md
```diff
@@ -68,7 +76,14 @@ Use the `warning` variant to provide feedback of a warning; users would still be
...
+- Error and Warning variants should not be dismissed
```
jraff: Why are we capitalizing Warning?

packages/Notification/Notification.md
```diff
@@ -1,8 +1,14 @@
 **Use the "Open isolated" button above to view this component in full-width mode.**
 
-The `Notification` spans the entire width of the screen, and aligns the message with the page content.
+A banner that highlights important messages.
 
-By default, notifications will be displayed in the `instructional` variant.
+Notifications are used to let users know:
+
+- System status such as an error or outage
+- Their actions has triggered a change such as confirmation of a payment
```
jraff: Change "has" to "have"

Result: "Their actions have triggered a change such as confirmation of a payment"

packages/Notification/Notification.md
```diff
@@ -1,8 +1,14 @@
...
+- Information such as letting the user know they have almost used all their data
+
+Colours help indicate the type of message. For errors use a red background. For other notifications use grey. Lastly, if the message is promotional use branded colours. These are reflected in code via the `variants` prop, and the background colour will change to match the desiered variant. For example, the `error` variant will produce a red background.
```
jraff: Change `variants` to `variant`

packages/Notification/Notification.md
```diff
@@ -17,6 +23,8 @@ By default, notifications will be displayed in the `instructional` variant.
   not use other complex components or paragraphs.
 - Use complete sentences, while keeping the content succinct.
 - Dismissible notifications should only be used for branded or general style notifications where there is no action needed from the user.
+- If the `variant` is `success` or `error`, there should be a `Box` with `between` level 3 between the icon and the text
```
jraff: I don't think this line should belong on our docs

packages/Notification/Notification.md
```diff
@@ -17,6 +23,8 @@ By default, notifications will be displayed in the `instructional` variant.
...
+- If the `variant` is `dismissible`, there should be a `Box` with `between` level 3 between the text and the dismiss button
```
jraff: I don't think this line should belong on our docs

packages/Notification/Notification.md
```diff
@@ -68,7 +76,14 @@ Use the `warning` variant to provide feedback of a warning; users would still be
 
 ### Dismissible
 
-Use the `dismissible` prop to allow users to dismiss the `Notification` at any time. The `dismissible` prop must be used together with the `dismissibleA11yLabel` to remain accessible. Once dismissed, the `Notification` is removed from the DOM.
+Notifications can either be passive or require action from the user. If the information is passive then it should be dismissible as indicated by an interactive close button on the far right. If the Notification requires an action then make the option(s) clear with a Chevron Link or Paragraph link style.
+
+Use the dismissible prop to allow users to dismiss the Notification at any time. The `dismissible` prop must be used together with the `dismissibleA11yLabel` to remain accessible. Once dismissed, the Notification is removed from the DOM.
```
jraff: Can we add backticks around dismissible?

packages/Notification/Notification.md
```diff
@@ -17,6 +23,8 @@ By default, notifications will be displayed in the `instructional` variant.
   not use other complex components or paragraphs.
 - Use complete sentences, while keeping the content succinct.
```
jraff: "When there are multiple notifications, the messages should be served top to bottom from most to least important. Remove any unnecessary information such as promotional messages that add to the noise"

Should be added after this line

packages/Notification/Notification.md
marcod1419: Instructional is a variant type though. It's not currently deprecated, should it be?

packages/Notification/Notification.md
marcod1419: I decided to capitalize all variant types. I'm flat hand on keeping that though

packages/Notification/Notification.md
marcod1419: This was part of the misunderstanding of this ticket, I'll remove it.

packages/Notification/Notification.md
jraff: My mistake.

Issue #1372, opened 2020-01-27
<ExpandCollapse.Panel /> teritaryText going on 2 lines in Safari browser


Jeffrey-Chang:
## Description

Not sure which characters and/or how long of a text has to be that will cause this specific bug but I've only noticed it when there is a `-$7.50` in the `tertiaryText` prop. Adding that specific text will cause `-` to be on one line and then `$7.50` will be on another line. This will _*ONLY*_ happen in `Safari` browser but _*DOES NOT*_ happen in `Chrome` and `Firefox`. I have also not tested this in `Microsoft Edge` and `Internet Explorer`.

## Reproduction Steps

1. Open the `Safari` browser
2. Head over to the main component link on the TDS website
3. Go to the `ExpandCollapse` section
4. In any of the `<ExpandCollapse.Panel />` add a `tertiaryText="-$7.50"`
5. Notice how the `-` will be on one line and then `$7.50` will be on another line

## Workaround details

N/A

## Recommendation

N/A

## Meta

- TDS component version: @tds/core-expand-collapse <- Not sure how far back it goes
- Willing to develop solution: No
- High impact: No

## Screenshots

![Screen Shot 2020-01-27 at 10 23 27 AM](https://user-images.githubusercontent.com/30445300/73187226-190f0e80-40ef-11ea-8bc9-a1527b49b3c6.png)

jraff:
@Jeffrey-Chang Issue is fixed in [@tds/core-expand-collapse@2.0.33](https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-expand-collapse%402.0.33)

PR #1387, opened 2020-02-11
fix(core-expand-collapse): prevent tertiary text overflow

marcod1419:
See #1372 

This prevents an issue where tertiary text will overflow onto a second line if it's too long.

### DIFF
--- a/packages/ExpandCollapse/PanelWrapper/PanelWrapper.jsx
+++ b/packages/ExpandCollapse/PanelWrapper/PanelWrapper.jsx
@@ -62,11 +62,13 @@ const ShowFromMd = styled.div({
   display: 'none',
   ...media.from('md').css({
     display: 'inline-block',
+    whiteSpace: 'nowrap',
   }),
 })
 
 const ShowUntilMd = styled.div({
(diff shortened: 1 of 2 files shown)


packages/ExpandCollapse/PanelWrapper/PanelWrapper.jsx
```diff
@@ -62,11 +62,13 @@ const ShowFromMd = styled.div({
   display: 'none',
   ...media.from('md').css({
     display: 'inline-block',
+    'white-space': 'nowrap',
```
jraff: Could this be written as `whiteSpace`?

packages/ExpandCollapse/PanelWrapper/PanelWrapper.jsx
```diff
@@ -62,11 +62,13 @@ const ShowFromMd = styled.div({
...
   }),
 })
 
 const ShowUntilMd = styled.div({
   display: 'inline-block',
+  'white-space': 'nowrap',
```
jraff: Could this be written as whiteSpace?

PR #1390, opened 2020-02-12
fix(shared-video-splash): simplify box usage in big play button

marcod1419:
This is a PR that addresses the following TODO in the Video Splash shared component:

`// TODO: After Box moves to Styled Components, use one Box with the between prop for spacing`

All items in the big play button are now wrapped by one box. As a result, the spacing for the items in the button is more centred. (As it probably should have been to begin with) A designer should look at this during QA along with a dev.


jraff:
Design should remain as is. We will not be moving forward with this cleanup right now.

PR #1396, opened 2020-02-19
Render Box between 0 as display block

jraff:
## Description

- Render `<Box between={0}>` as `display: block`

- [x] New code is unit tested

### DIFF
--- a/packages/Box/Box.jsx
+++ b/packages/Box/Box.jsx
@@ -43,7 +43,7 @@ const inlineBetweenStyles = props =>
     { between: props.between, inline: props.inline },
     ({ between, inline }, breakpoint) => {
       const base = {
-        display: between || !inline ? 'flex' : 'block',
+        display: typeof between !== 'undefined' ? 'flex' : 'block',
         flexDirection: inline ? 'row' : 'column',
       }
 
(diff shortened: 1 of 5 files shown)


packages/Box/Box.jsx
```diff
@@ -43,7 +43,7 @@ const inlineBetweenStyles = props =>
     { between: props.between, inline: props.inline },
     ({ between, inline }, breakpoint) => {
       const base = {
-        display: between ? 'flex' : 'block',
+        display: between || !inline ? 'flex' : 'block',
```
theetrain: Let's update this to `display: between !== undefined ? 'flex' : 'block'`

We want it to be `display: flex` whenever `between` is defined. Currently, when `between={0}`, `0` is coerced as a falsy value.

packages/Box/Box.jsx
```diff
@@ -43,7 +43,7 @@ const inlineBetweenStyles = props =>
     { between: props.between, inline: props.inline },
     ({ between, inline }, breakpoint) => {
       const base = {
-        display: between ? 'flex' : 'block',
+        display: typeof between !== 'undefined' ? 'flex' : 'block',
```
marcod1419: Could we not just do `display: between !== undefined ? ‘flex’ : ‘block’`? (Drop the `typeof`)

PR #1429, opened 2020-04-01
Add Open Carousel

marcod1419:
This PR adds the "Open" variant for Carousel. E2E tests and unit tests will be expanded upon once the "Card" variant is in development. Additionally, docs are still awaiting assets, so the sandbox content is not final.

### DIFF
--- a/packages/ContentCarousel/ContentCarousel.jsx
+++ b/packages/ContentCarousel/ContentCarousel.jsx
@@ -1,0 +1,252 @@
+import React, { useState, useEffect } from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import anime from 'animejs'
+import { useDrag } from 'react-use-gesture'
+
+import { media } from '@tds/core-responsive'
+import { safeRest } from '@tds/util-helpers'
+import NavButton from './NavButton/NavButton'
+import Item from './Item/Item'
+import PageIndicator from './PageIndicator/PageIndicator'
+
+/**
+ * @version ./package.json
+ */
+
+const CarouselContainer = styled.div({
+  width: '100%',
+  height: '100%',
+  position: 'relative',
+  overflow: 'hidden',
+})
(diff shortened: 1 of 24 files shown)


packages/ContentCarousel/ContentCarousel.jsx
```diff
@@ -0,0 +1,252 @@
...
+          handleSlideTransition('left', 1)
+        } else if (movement[0] > 0 && currentPage > 1) {
+          handleSlideTransition('right', -1)
+        }
+      }
+    },
+    { filterTaps: true, axis: 'x', threshold: 100 }
+  )
+
+  const preloadImage = url => {
```
jraff, line 184: Out of curiosity, what do the `preloadImage()`, `findImages()`, and the `useEffect()` do?

marcod1419: These functions load the `picture` prop of every slide on initial render. It stops images from loading in only on slide switch.

packages/ContentCarousel/PageIndicator/PageIndicator.jsx
```diff
@@ -0,0 +1,107 @@
...
+  })
+  return (
+    <Dot onClick={onClick} isCurrentPage={isCurrentPage} aria-current={isCurrentPage}>
+      <FocusOutline />
+    </Dot>
+  )
+}
+
+const PageIndicator = ({ currentPage, totalPages, changePage, handleSlideSkip, ...rest }) => {
+  const generateDots = () => {
```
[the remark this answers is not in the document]
marcod1419: I think memoization is unnecessary. The page indicator will frequently change props as the user navigates the component, making the gains from using `React.memo` pretty marginal IMO. `generateDots` is in a function just so we don't have a lot of logic in the component's JSX.

theetrain:
A new PR will be made.

Issue #1441, opened 2020-04-28
TDS Video component - event tracking


mike-bunce:
## Description

We need to be able to track events, play event primarily so that we can have analytics tracking on videos with TDS Video component.

## Recommendation

Add a callback we can access which provides with access to the event object.

## Meta

- TDS component: @tds/core-video
- Willing to develop solution: Yes, if need be
- Has workaround: No
- High impact: Yes, we cannot track video analytics for our campaign, so we don't know the effectiveness of the video we have pubilshed.
- Designer/Developer Pair: Mike Bunce, no designer required.

theetrain:
This enhancement sounds good to us and we welcome pull requests.

The idea is to add a new `eventCallback` prop that users of core-video could use to pass a handler function accepting an `customerEvents` object parameter that contains customer events such as when the play button is clicked, or how much time has elapsed from the video.

jraff:
Thank you for the contribution @mike-bunce. This feature has been updated with [@tds/core-video@1.3.0](https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-video%401.3.0)

PR #1445, opened 2020-05-09
Adding analytics to video

mike-bunce:
## Related issues

https://telusdigital.atlassian.net/browse/RL-3649 
#1441 

## Description

Adds tracking for analytics - You can pass in a callback function to Video, in order to track analytics. The parameter this function takes will give you access to the analytics object which will be updated every time an action is fired (this inclues "play", "pause", percentage watched as an interval of "25%")

- [x] New code is unit tested
- [x] Commits follow our [Developer Guide](https://tds.telus.com/contributing/developer-guide.html#make-a-commit)

### DIFF
--- a/packages/Video/Video.jsx
+++ b/packages/Video/Video.jsx
@@ -102,6 +102,7 @@ class Video extends React.Component {
       captionsMenuOpen: false,
       isMobile: false,
       videoPlayerWidth: 0,
+      percentPlayed: 'watched 0%',
     }
   }
 
@@ -168,6 +169,9 @@ class Video extends React.Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
+    if (nextState.videoIsPlaying !== this.state.videoIsPlaying) {
+      this.updateAnalyticsData()
+    }
     if (nextState !== this.state || nextProps !== this.props) {
       return true
     }
@@ -177,6 +181,7 @@ class Video extends React.Component {
   componentWillUnmount() {
     this.clearInactivityTimer()
     window.onresize = undefined
+    clearInterval(this.state.intervalId)
   }
 
   // ******** Begin Initialization Functions *********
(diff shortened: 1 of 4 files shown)


packages/Video/Video.jsx
```diff
@@ -311,6 +316,44 @@ class Video extends React.Component {
...
+    }
+    if (percentValue < 100) {
+      return this.setState({ percentPlayed: 'watched 75%' })
+    }
+    if (percentValue === 100) {
+      return this.setState({ percentPlayed: 'watched 100%' })
+    }
+    return false
+  }
+
```
jraff, line 356: Would it be better to return `elapsedTime` and `totalTime` instead of `percentPlayed`. This way outcome teams can implement their own interpretation of the metrics.

packages/Video/Video.md
```diff
@@ -21,6 +21,10 @@ Each video will appear in the "Quality" dialogue from top to bottom sorted by `q
 
 Tracks are supplied to the player through the `tracks` prop, which accepts an array of objects. Each object in the array represents a track file. Typically, these will be different languages or separate files for closed captions, subtitles, or descriptions. If no tracks are provided, the captions button will not be visible in the player.
 
+#### Analytics tracking
```
jraff, line 24: Would it be best for the `analyticsObject` to include all possible metrics at every callback? This way we can add props to the analytics object over time and avoid breaking changes. I recommend returning a full object at every callback.

```
// Customer clicks play
analyticsObject = {
  state: 'play',
  elapsedTime: 0,
  totalTime: 60
}

// Customer clicks pause after 20 seconds
analyticsObject = {
  state: 'pause',
  elapsedTime: 20,
  totalTime: 60
}
```

packages/Video/Video.jsx
mike-bunce: Hi Jordan, I can change percentPlayed to elapsedTime no problem. There's no need for a totalTime, because "videoLength" already exists.

packages/Video/Video.md
mike-bunce: So, I have been requested by the analytics team to return an object that contains "name", "action" and "details". Name, and details we are going to handle on the application side, but action we need from the component and it has a very specific format (which is what we are returning in the component). It needs to have contain a string such as:
"action": "play", // or "pause", "watched 25%", etc.
So I am following that format, and my understanding is that this should work for all teams and it should match the analytics object that the web videos return (from YouTube)

packages/Video/Video.jsx
```diff
@@ -311,6 +316,44 @@ class Video extends React.Component {
...
+    if (previousValue !== this.state.elapsedTime) {
+      const analyticsObject = {}
+      analyticsObject.action = this.state.elapsedTime
+      this.props.analyticsTracking(analyticsObject)
+    }
+  }
+
+  percentageBucket = percentValue => {
+    if (percentValue < 25) {
+      return this.setState({ elapsedTime: 'watched 0%' })
```
theetrain, line 340: You can compare previous values when you pass a function into `setState`.

packages/Video/Video.jsx
```diff
@@ -837,6 +887,7 @@ Video.defaultProps = {
   crossOrigin: undefined,
   simpleMode: false,
   videoBorder: false,
+  analyticsTracking: () => {},
```
theetrain, line 890: This can be `undefined` by default.

packages/Video/Video.jsx
```diff
@@ -311,6 +316,44 @@ class Video extends React.Component {
...
+  // required for analytics
+  calculatePercentageWatched = () => {
+    let percentValue = (this.state.videoCurrentTime / this.state.videoLength) * 100
+    percentValue = Math.round(percentValue)
+    const previousValue = this.state.elapsedTime
+    this.percentageBucket(percentValue)
+    if (previousValue !== this.state.elapsedTime) {
+      const analyticsObject = {}
+      analyticsObject.action = this.state.elapsedTime
+      this.props.analyticsTracking(analyticsObject)
```
theetrain, line 334: Also include the `name` and `details` keys.

mike-bunce:
Embargo approval: https://services.trlcloud.com/prod/requestStatus/index.html?id=367645572

PR #1457, opened 2020-07-13
Adding a prop as an option to make the button full width

sudhirraivns0103:
Hi @mike-bunce ,
I have added prop as on option to make the button full width.

### DIFF
--- a/packages/Button/Button.jsx
+++ b/packages/Button/Button.jsx
@@ -76,7 +76,10 @@ export const StyledButton = styled.button(
   ({ fullwidth }) => {
     let width
     if (fullwidth) {
-      width = '100% !important'
+      return {
+        width: '100%',
+        ...media.from('md').css({ width: '100%' }),
+      }
     }
     return {
       width,
--- a/packages/Button/Button.jsx
+++ b/packages/Button/Button.jsx
@@ -18,7 +18,6 @@ const preventDisabling = ({ disabled, ...props }) => {
 
   return props
 }
-const fullwidthButton = '100% !important'
 const baseButton = {
   margin: 0,
   padding: '0 2rem',
@@ -77,7 +76,7 @@ export const StyledButton = styled.button(
   ({ fullwidth }) => {
     let width
     if (fullwidth) {
-      width = fullwidthButton
+      width = '100% !important'
     }
     return {
       width,
--- a/packages/Button/Button.jsx
+++ b/packages/Button/Button.jsx
@@ -18,7 +18,7 @@ const preventDisabling = ({ disabled, ...props }) => {
 
   return props
 }
-
+const fullwidthButton = '100% !important'
 const baseButton = {
   margin: 0,
   padding: '0 2rem',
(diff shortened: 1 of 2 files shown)


packages/ButtonLink/ButtonLink.jsx
```diff
@@ -87,6 +100,7 @@ ButtonLink.propTypes = {
    * Target URL
    */
   href: PropTypes.string,
+  fullwidth: PropTypes.string,
```
theetrain, line 103: Please change this PropType to a boolean, and add comments above such as:

```jsx
/**
 * If `true`, sets `ButtonLink` to 100% width.
 * @since 2.2.0
 */
fullWidth: PropTypes.bool,
```

packages/ButtonLink/ButtonLink.jsx
```diff
@@ -97,6 +111,7 @@ ButtonLink.defaultProps = {
   reactRouterLinkComponent: null,
   to: null,
   href: null,
+  fullwidth: null,
```
theetrain, line 114: Default should be `false`

sudhirraivns0103: Hi @theetrain,
1 - I have added comment in buttonLink component. 
2 -I have added prop as on option to make the button full width for core-button component.

packages/Button/Button.jsx
```diff
@@ -74,6 +73,15 @@ export const StyledButton = styled.button(
       color,
       '&:hover': hover,
     }
+  },
+  ({ fullwidth }) => {
+    let width
+    if (fullwidth) {
+      width = fullwidthButton
```
theetrain, line 80: Here you can add `width = '100%'` and delete the variable on L21.

packages/Button/Button.jsx
```diff
@@ -18,7 +18,7 @@ const preventDisabling = ({ disabled, ...props }) => {
 
   return props
 }
-
+const fullwidthButton = '100% !important'
```
theetrain, line 21: This can be removed.

packages/Button/Button.jsx
sudhirraivns0103: Hi @theetrain,
I can use 'width = '100%' !important' because parent defined width : auto.

packages/Button/Button.jsx
sudhirraivns0103: Hi @theetrain,
I have no permission to code commit on tds-core repository
I am getting error :
remote: Permission to telus/tds-core.git denied to sudhirraivns0103.
fatal: unable to access 'https://github.com/telus/tds-core/': The requested URL returned error: 403

packages/Button/Button.jsx
sudhirraivns0103: I have removed variable

sudhirraivns0103:
Hi @theetrain,
I have removed variable and !important key from width.

theetrain:
This has been replaced with #1458

PR #1463, opened 2020-07-25
feat(core-button-link):  adding a prop to make the button full width

sudhirraivns0103:
I have added prop as on option to make the button full width.
Closes #1461


packages/ButtonLink/ButtonLink.jsx
```diff
@@ -43,6 +43,20 @@ const StyledButtonLink = styled(StyledButton)(
         color: hoverColor,
       },
     }
+  },
+  ({ fullWidth }) => {
+    let width
+    if (fullWidth === 'true') {
```
mike-bunce, line 49: Can you write this as a ternary statement instead?

e2e/visual/components/CartesianButtonLink.jsx
```diff
@@ -12,6 +12,7 @@ const CartesianButtonLink = props => (
     variant={['primary', 'secondary', 'inverted']}
     href="http://tds.telus.com"
     children="TELUS Design System"
+    fullWidth="false"
```
mike-bunce, line 15: Can you write this as requested: fullWidth={[false,true]}

packages/ButtonLink/ButtonLink.jsx
sudhirraivns0103: Hi @mike-bunce,
I have updated PR again with requested changes

packages/ButtonLink/ButtonLink.jsx
```diff
@@ -87,6 +98,11 @@ ButtonLink.propTypes = {
    * Target URL
    */
   href: PropTypes.string,
+  /**
+   * If `true`, sets `ButtonLink` to 100% width.
+   * @since 2.2.0
+   */
+  fullWidth: PropTypes.oneOf(['true', 'false']),
```
jraff, line 105: Can we use a boolean instead of true/false strings?

sudhirraivns0103: Hi @jraff,
I have changed boolean instead of true/false

packages/ButtonLink/ButtonLink.jsx
```diff
@@ -43,6 +43,14 @@ const StyledButtonLink = styled(StyledButton)(
         color: hoverColor,
       },
     }
+  },
+  ({ fullWidth }) => {
+    const width = fullWidth === 'true' ? '100%' : 'auto'
```
jraff, line 48: This line should be changed to `const width = fullWidth ? '100%' : 'auto'` as `fullWidth` should no longer be a string.

sudhirraivns0103: Hi @jraff,
I have update your requested change.

PR #1471, opened 2020-08-07
Full bleed image card

meeschka:
## Card with Full Bleed Image

Creates an optional prop called fullBleedImage in the card component, which can be used to add an Image to the Card without margins. Spacing properties passed to the Card are instead applied only to the children content, not the image. The image position (top, down, left, right, or none) is responsive for maximum flexibility.

An example of a card with a full bleed image at the right:
Screen Shot 2020-08-07 at 11 55 46 AM
 
A responsive fullHeight card, with a full bleed image on desktops and none on mobile:
Screen Shot 2020-08-07 at 11 55 57 AM
Screen Shot 2020-08-07 at 11 56 52 AM

The card image does not distort due to changes in card size. These extra paragraphs were added after, but the image remains the same:
Screen Shot 2020-08-12 at 12 53 21 PM


packages/Card/Card.jsx
```diff
@@ -71,9 +74,35 @@ const Card = ({ variant, children, fullHeight, spacing, ...rest }) => {
...
+      return styles
+    })
+
+  const StyledDiv = styled.div(ContainerStyles)
   return (
-    <StyledCard {...safeRest(rest)} fullHeight={fullHeight} variant={variant} {...spacingProps}>
-      {children}
+    <StyledCard {...safeRest(rest)} fullHeight={fullHeight} variant={variant}>
+      <StyledDiv {...fullBleedImage}>
+        {fullBleedImage ? <Image {...fullBleedImage} /> : ''}
```
janeszelag, line 103: We can change this to `{fullBleedImage && <Image {...fullBleedImage} />}`

packages/Card/Card.jsx
```diff
@@ -71,9 +74,35 @@ const Card = ({ variant, children, fullHeight, spacing, ...rest }) => {
     spacingProps.inset = 4
   }
 
+  const ContainerStyles = imageProps =>
+    handleResponsiveStyles({ position: imageProps?.position }, ({ position }) => {
```
canrozanes, line 78: We don't support ES2020 yet. Do you mind converting your optional chaining to an explicit check ie.
`imageProps && imageProps.position`

meeschka: Will do!

packages/Card/Card.jsx
```diff
@@ -71,9 +74,37 @@ const Card = ({ variant, children, fullHeight, spacing, ...rest }) => {
...
+      return styles
+    })
+
+  const StyledDiv = styled.div(ContainerStyles)
   return (
-    <StyledCard {...safeRest(rest)} fullHeight={fullHeight} variant={variant} {...spacingProps}>
-      {children}
+    <StyledCard {...safeRest(rest)} fullHeight={fullHeight} variant={variant}>
+      <StyledDiv {...fullBleedImage}>
+        {fullBleedImage && <Image {...fullBleedImage} />}
```
canrozanes, line 105: Can you please pass individual props instead of prop spreading?

There are two reasons I can think of:
1- Readability: When I just see `{...fullBleedImage}` I have no idea what exactly is being passed to `Image`.
2- Unnecessary prop passing: `Image` component by design doesn't take a `position` prop when you spread the `fullBleedImage` object, you end up passing an unnecessary position prop to it.

packages/Card/handleResponsiveStyles.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const handleResponsiveStyles = (props, styleFn) => {
+  const breakpoints = prepareArray(props)
+    .filter(bp => Object.keys(bp.props).length > 0)
+    .reduce(handleBoundaryCrossing, [])
+  return generateStyles(breakpoints, styleFn)
+}
+
+export default handleResponsiveStyles
```
canrozanes, line 37: I see that you copied this function `handleResponsiveStyles` from the Box component. It's better to define a function once and reuse it in multiple places than copy and paste it each time you want to use it.

That way, you end up having less code to maintain. For more info [read here](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

With that spirit, I suggest you move this function to `'@tds/util-helpers'` package and import it to both Card and Box components.

packages/Card/Card.jsx
```diff
@@ -71,9 +74,37 @@ const Card = ({ variant, children, fullHeight, spacing, ...rest }) => {
...
+        justifyContent: 'space-between',
+        '> img': {
+          display: position === 'none' ? 'none' : 'block',
+          margin: 'auto',
+        },
+      }
+      return styles
+    })
+
+  const StyledDiv = styled.div(ContainerStyles)
```
jraff, line 101: `styled-components` should be declared outside of any render method. In this case the `StyledDiv` should be declared outside of `Card`, just as `StyledCard` is.

See https://styled-components.com/docs/faqs#why-should-i-avoid-declaring-styled-components-in-the-render-method

packages/Card/Card.jsx
```diff
@@ -71,9 +74,37 @@ const Card = ({ variant, children, fullHeight, spacing, ...rest }) => {
...
   return (
-    <StyledCard {...safeRest(rest)} fullHeight={fullHeight} variant={variant} {...spacingProps}>
-      {children}
+    <StyledCard {...safeRest(rest)} fullHeight={fullHeight} variant={variant}>
+      <StyledDiv {...fullBleedImage}>
```
jraff, line 104: It seems as though we will be returning this `StyledDiv` regardless of whether or not the `fullBleedImage` prop is provided.

Can we instead check for `fullBleedImage` and then return this structure. Otherwise just return the existing `StyledCard` This should eliminate unnecessary DOM elements.

packages/Card/Card.jsx
```diff
@@ -71,9 +74,37 @@ const Card = ({ variant, children, fullHeight, spacing, ...rest }) => {
...
+        right: 'row-reverse',
+        top: 'column',
+        bottom: 'column-reverse',
+        none: 'row',
+      }
+      const styles = {
+        display: 'flex',
+        flexDirection: direction[position],
+        justifyContent: 'space-between',
+        '> img': {
```
jraff, line 93: I feel iffy of this `'> img'` selector.

If we can guarantee  that the only child of the `styled-component` will be the `fullBleedImage` `Image`  I would feel much more confident.

packages/Card/Card.jsx
meeschka: Makes sense, change implemented :)

packages/Card/Card.jsx
meeschka: Makes sense, thanks for the reference! I've fixed this.

packages/Card/Card.jsx
meeschka: Makes sense, change implemented

packages/Card/Card.jsx
meeschka: We can guarantee that it will only select the full bleed image. This is a direct descendent selector, and there are only two direct descendents: the full bleed image, and a box with all the children. There will never be another image direct descendent in this div.

jraff:
This is ready for release.

jraff (approved):

Issue #1503, opened 2020-10-28
Strikethrough option in PriceLockup


elissa-matsushita:
## Description

- On the Optik TV package cards, we need to update the cards to have price lockup with strike-through - medium). 

## Recommendation

- Add strikethrough prop to PriceLockup

## Designs

- Screenshots:
<img width="282" alt="Screen Shot 2020-10-27 at 8 50 01 PM" src="https://user-images.githubusercontent.com/66031253/97388767-fe441300-1895-11eb-880f-10968bf63f23.png">

- Link to Invision: https://telus.invisionapp.com/d/main?origin=v7#/console/20296211/433873122/preview?scrollOffset=0

## Meta

- TDS component: PriceLockup 
- Willing to develop solution: Yes
- Has workaround: No 
- High impact: No 
- Dev: @elissa-matsushita , Designer: @twirlingsky

twirlingsky:
additional requirements: 

- Will need the French version as well
- the strike-through is purely a visual solve, it MUST be accompanied by an aria tag.  "was $xx"

twirlingsky:
DSM guidelines: https://telus.invisionapp.com/dsm/telus/tds-community/asset/components/5f8f75024d7a1163f43fc8ec

jraff:
Fixed with https://github.com/telus/tds-core/releases/tag/%40tds%2Fcore-price-lockup%402.2.0

PR #1505, opened 2020-11-05
fix(core-price-lockup): use a11ycontent instead of aria label

elissa-matsushita:
screen readers do not read aria labels on non-interactive elements. we will use a11yContent component instead

## Related issues

See #1503 

## Description

- realized that screen readers do not read aria labels on non-interactive elements, so we are now using the `A11yContent` component wrap the descriptive text instead

- [x] New code is unit tested


jraff (changes requested):
Please delete `packages/PriceLockup/package-lock.json`. We don't commit package-lock.json in components.

jraff (approved):

PR #1517, opened 2021-01-12
Icon size scaling with browser font size

jraff:
## Description

- Allow icon size scaling with browser font size

- [x] New code is unit tested

### DIFF
--- a/packages/FeedbackIcon/FeedbackIcon.jsx
+++ b/packages/FeedbackIcon/FeedbackIcon.jsx
@@ -1,13 +1,19 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import styled from 'styled-components'
+
 import { getCopy, safeRest } from '@tds/util-helpers'
 
 import { warn } from '../../shared/utils/warn'
 
+const StyledSVG = styled.svg(({ width, height }) => ({
+  width: `${width}rem`,
+  height: `${height}rem`,
+}))
 /**
  * @version ./package.json
  */
-const FeedbackIcon = ({ copy, copyDictionary, optionalText, children, ...rest }) => {
+const FeedbackIcon = ({ width, height, copy, copyDictionary, optionalText, children, ...rest }) => {
   if (rest.onClick) {
     warn(
       'FeedbackIcon',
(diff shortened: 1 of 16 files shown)


packages/FeedbackIcon/FeedbackIcon.jsx
```diff
@@ -25,10 +31,16 @@ const FeedbackIcon = ({ copy, copyDictionary, optionalText, children, ...rest })
   }
 
   return (
-    <svg {...safeRest(rest)} role="img" aria-hidden={a11yText === '' ? true : undefined}>
+    <StyledSVG
+      {...safeRest(rest)}
+      role="img"
+      aria-hidden={a11yText === '' ? true : undefined}
+      width={`${width / 16}rem`}
+      height={`${height / 16}rem`}
```
canrozanes, line 39: what is the the significance of 16 here? Can we store that as a const `MAGIC_NUMBER = 16` (with a more descriptive variable name of course) on the top of the file and use it here?

packages/InteractiveIcon/Limited.jsx
```diff
@@ -11,8 +13,8 @@ import StyledInteractiveIconSVG from './shared/StyledInteractiveIconSVG'
 
 const iconSize = props =>
   handleResponsiveStyles({ size: props.size }, ({ size }) => ({
-    width: size === 20 ? '20px' : '24px',
-    height: size === 20 ? '20px' : '24px',
+    width: size === 20 ? '1.25rem' : '1.5rem',
+    height: size === 20 ? '1.25rem' : '1.5rem',
```
canrozanes, line 17: Is 1.25rem and 1.5rem magic numbers or is this a TDS standard? I can't really tell here from looking at the code.

packages/FeedbackIcon/FeedbackIcon.jsx
jraff: 16px is the default font size in almost all browsers, and TDS assumes this. This pixel-to-rem conversion could be something that we move into sometime like`@tds/shared-utils` as it is currently being used in a couple other different places.

packages/InteractiveIcon/Limited.jsx
jraff: 1.25 rem and 1.5 rem are very common to see in TDS (see https://github.com/telus/tds-core/blob/master/packages/Box/Box.jsx#L9). However this could be another place where we could use something like a pixel-to-rem conversion from the `@tds/shared-utils` package.

canrozanes (approved):

PR #1520, opened 2021-01-29
fix(core-button): no hover on touch screens

cianfoley-nearform:
## Related issues

Global issue

## Description

On touch displays, after a button is clicked it remains in the hover state.

We use hover media query here to remove hover styles on screens that don't handle hover well.

We cannot use `@media (hover: hover)` because ie11 would not render the hover style, so we enable by default and reset the style using `@media (hover: none)

### DIFF
--- a/packages/Button/Button.jsx
+++ b/packages/Button/Button.jsx
@@ -128,6 +128,13 @@ const getVariant = ({ variant, rank }) => {
     border,
     transition,
     '&:hover': hover,
+    '@media (hover: none)': {
+      '&:hover': {
+        boxShadow: 'none',
+        backgroundColor,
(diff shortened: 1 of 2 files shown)


jraff:
LGTM, please fix failing GitHub actions and we can proceed with QA.

cianfoley-nearform:
Thanks @jraff, should I be seeing a package version update in the TDSBot comment above?

iamstuartwilson (dismissed):
Neat way of doing it! 👏

cianfoley-nearform:
@jraff, what are the next steps here to get into QA?

jraff (approved):

Issue #1485, opened 2020-09-10
TermsAndConditions component throughs unique key warning for indexedContent and nonIndexedContent props of type Node


manjitsingh88:
## Screenshots

<img width="1321" alt="Screen Shot 2020-09-10 at 5 44 30 PM" src="https://user-images.githubusercontent.com/70860252/92814105-993d6800-f390-11ea-8493-fb76aea9e290.png">

## Description

- Tech Health
- TermsAndConditions component throughs unique key warning for `indexedContent` and `nonIndexedContent` props of type Node
- n number of console errors as each page has terms and conditions.

## Reproduction Steps

1. View the `TermsAndConditions` component
2. Click on any link with terms and conditions (In dev mode only)
3. Observe react console warnings

## Recommendation

- check if `indexedContent` or `nonIndexedContent` prop is of type object and has props property.
- stringify the react node props object and assign it as string key.
JSON.stringify(node?.props)

## Meta

- TDS component version: @tds/core-terms-and-conditions
- Willing to develop solution: Yes
- High impact: No

PR #1557, opened 2021-04-06
fix(core-terms-and-conditions): use index as key when mapping

kyletsang:
## Related issues

See #1485

## Description

Fixes key warning by using an index since `indexedContent` and `nonIndexContent` allows components

Unfortunately having a number of issues running the environment - gitbook doesn't install.  Hoping this change is minor enough to submit?


kyletsang:
Unable to update snapshots on my end.  I believe a number of outdated dependencies are causing issues with my version of Node (14)

jraff (approved):

jraff (changes requested):
Please fix the unit tests.

jraff (approved):

PR #1907, opened 2023-03-22
TDS sunset warning added for Core components and Gitbook pages

shahzaibkhalid:
This PR adds TDS sunset warning for Core components and Gitbook pages. It's straightforward to do that for core components pages as those are based on StyleGuidist but the rest of the site is based on GitBooks which doesn't have any firest-class way of doing it. So I had to add a gitbook plugin to inject HTML dynamically.

SCR-20230322-jmqf


srikanthkh (approved):
lgtm 🚀

